PR #5, opened 2022-09-16
Add building blocks for Acl plugin

mooori:
# Acl: Access control lists
This PR adds building blocks for an Acl plugin, inspired by [OpenZeppelin Access Control](https://docs.openzeppelin.com/contracts/4.x/api/access).

There is a contract in `near-plugins/tests/contracts/access_controllable` which shows how to use the plugin. In integration test `near-plugins/tests/access_controllable.rs` the contract is compiled, deployed and transactions are sent to it (all via `workspaces-rs`).

```
# Command to execute the tests:
cargo test --test access_controllable
```

__Note__: If this doesn't compile, #6 should fix it.

# Overview
- The contract developer specifies an enum whose variants represent roles.
- The developer derives `AccessControlRole` for that enum. It defines a bitflag type that encodes permissions and implements helpers for the conversion between roles (i.e. variants of the enum) and bitflags.
- The contract itself gets the attribute `#[access_control]` which injects the structure that manages permissions into the contract.
- The developer can then use functions and attributes to interact with Acl, e.g.
  - `self.acl_has_role(role, account_id)`)
  - `#[access_control_any(roles(Role::LevelA, Role::LevelB)])`

# Roadmap
This PR adds the basics for Acl and the contract and integration test mentioned above ensure that everything is wired together correctly. Upcoming PRs will add functionality to make this plugin feature equivalent to OpenZeppelin Access Control.

The target branch is `acl` (instead of master) since there will be more PRs on top of this.


near-plugins/src/access_control_role.rs
```diff
@@ -0,0 +1,10 @@
+pub trait AccessControlRole {
+    /// Returns the bitflag corresponding to the super admin permission.
+    fn acl_super_admin_permission_bitflag() -> u128;
+
+    /// Returns the bitflag corresponding to the admin permission for the role.
+    fn acl_admin_permission_bitflag(self) -> u128;
+
+    /// Returns the bitflag corresponding to the role's permission.
+    fn acl_permission_bitflag(self) -> u128;
```
mfornet, line 9: What do you think about removing the suffix `_bitflag`

Tbh, I don't expect user to be using these functions directly most of the time, so doesn't really matter much.

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+use std::convert::TryFrom;
+
+#[derive(AccessControlRole, BorshSerialize, BorshDeserialize)]
+pub enum Role {
+    Level1,
+    Level2,
+    Level3,
+}
+
+#[access_control(storage_prefix = "__foo", role_type = "Role")]
```
mfornet, line 14: Would this work if you replace `role_type = "Role"` with `Role`

```suggestion
#[access_control(Role)]
```

Omitted `storage_prefix` for clarity. I understand you don't have access to the trait per-se, but you only need access to the name of the trait.

----

With regards to the `storage_prefix` let's use some sensible default like: `__ACL` (+ allow overriding)

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    // The contract can interact with Acl by:
+    //
+    // a) Calling functions, e.g.
+    //    self.acl.has_role(role, account_id)
+    //
+    // b) Using attributes (not yet implemented), e.g.
+    //    #[acl_any(Role::Level1, Role::Level2)]
+    //    pub fn foo(&mut self) {}
```
mfornet, line 42: Once this function is implemented, recommended tests:

- Account with no role should fail
- Account with role level 1 should succeed
- Account with role level 2 should succeed
- Account with role level 3 should fail
- Account with roles level 1 and level 3 should succeed
- Account with role level 1 Admin should fail
- Account with role super admin should fail

near-plugins/src/access_control_role.rs
mooori: True with `_bitflag` it's quite verbose. Removed in 2793779.

near-plugins/tests/contracts/access_controllable/src/lib.rs
mooori: > Would this work if you replace role_type = "Role" with Role

I'll try if this is possible with `darling` (which handles argument parsing).

> With regards to the storage_prefix let's use some sensible default like: __ACL (+ allow overriding)

That's implemented :)

```rust
// Can't find a way to insert a link to a new file, so just pasting the code from
// near-plugins-derive/src/access_controllable.rs L49-51
let storage_prefix = macro_args
    .storage_prefix
    .unwrap_or_else(|| DEFAULT_STORAGE_PREFIX.to_string());
```

Changes in e8dfbac:
- Omit `storage_prefix`. Let's only specify _required_ arguments in this contract?
- Then we could add another (very simple) contract in which _all optional_ arguments are set

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+            return TokenStream::from(e.write_errors());
+        }
+    };
+    let roles = macro_args.roles;
+    assert!(roles.len() > 0, "Specify at least one role");
+
+    let acl_check = quote! {
+        let __roles: Vec<&str> = vec![#(#roles.into()),*];
+        let __roles_ser: Vec<String> = __roles.iter().map(|&role| role.into()).collect();
+        let __account_id = ::near_sdk::env::predecessor_account_id();
```
mfornet, line 268: There could be some potential name conflicts in scope if several macros try to create `__account_id`, what do you think about using `__acl_account_id` instead?

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+        if !self.acl_has_any_role(__roles_ser, __account_id) {
```
mfornet, line 269: Add TODO for a future revision that handles the case where a single role is passed so that it is treated simpler (no vector manipulation, and call directly to `acl_has_role`). I consider it relevant because I expect this to be the most common case.

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+    let bitflags_type = new_bitflags_type_ident(Span::call_site());
+    if let Err(e) = inject_acl_field(&mut input, &acl_field, &acl_type) {
+        return TokenStream::from(e.to_compile_error());
+    }
+    let ItemStruct { ident, .. } = input.clone();
+
+    let macro_args = match MacroArgs::from_list(&attr_args) {
+        Ok(args) => args,
+        Err(e) => {
+            return TokenStream::from(e.write_errors());
```
mfornet, line 39: Why using `write_errors` versus `to_compile_error`?

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+                self.has_any_permission(target, account_id)
+            }
+
+            fn has_any_permission(&self, target: #bitflags_type, account_id: &::near_sdk::AccountId) -> bool {
+                let permissions = self.get_or_init_permissions(account_id);
+                target.intersects(permissions)
+            }
+
+            /// Adds `account_id` to the set of `permission` bearers.
+            fn add_bearer(&mut self, permission: #bitflags_type, account_id: &::near_sdk::AccountId) {
```
mfornet, line 167: I want to highlight that at this point, we assume that `permission` will contain **exactly** one bit active. I checked, and this is the case for the only usage of `add_bearer,` but the function is exposed publicly (i.e, developers can call `self.acl.add_bearer(...)` themselves. We should consider adding this check.

```
fn is_power_of_2(num: u128) -> bool {
   num > 0 && ((num & (num - 1)) == 0)
}
```

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+        #[near_bindgen]
+        impl AccessControllable for #ident {
+            fn acl_storage_prefix() -> &'static [u8] {
+                (#storage_prefix).as_bytes()
+            }
+
+            #[private]
+            fn acl_grant_role_unchecked(&mut self, role: String, account_id: ::near_sdk::AccountId) -> bool {
+                let role = <#role_type>::try_from(role.as_str()).expect(#ERR_PARSE_ROLE);
+                self.#acl_field.grant_role_unchecked(role, &account_id)
```
mfornet, line 192: Can you remind me what was the reason to have acl as an injected field versus having a function that builds this object dynamically? For example:

```suggestion
                self.__acl_object().grant_role_unchecked(role, &account_id)
```

mfornet: The part I don't like about injecting the code without letting developers know is that they might struggle to upgrade their contracts and serialize/deserialize the state later.

near-plugins-derive/src/access_controllable.rs
```diff
@@ -0,0 +1,287 @@
...
+                    .iter()
+                    .map(|role| {
+                        <#bitflags_type>::from_bits(role.acl_permission())
+                            .expect(#ERR_PARSE_BITFLAG)
+                    })
+                    .collect();
+                let target = permissions.iter().fold(
+                    <#bitflags_type>::empty(),
+                    |acc, &x| acc | x,
+                );
```
mfornet, line 157:
```suggestion
                let target = roles
                    .iter()
                    .map(|role| {
                        <#bitflags_type>::from_bits(role.acl_permission())
                            .expect(#ERR_PARSE_BITFLAG)
                    }).fold(
                        <#bitflags_type>::empty(),
                        |acc, &x| acc | x,
                    );
```

mfornet (changes requested):
The code looks good. Please address the comments above and let's merge it. Then we need to add admin primitives and public functions.

near-plugins/tests/contracts/access_controllable/src/lib.rs
mooori: The first 5 cases are covered in this PR. Admin related functionality will be added in a follow up PR. There's a TODO in the code to then add the last two cases.

near-plugins/tests/access_controllable.rs
```diff
@@ -0,0 +1,254 @@
...
+    method_name: String,
+    /// The roles that are allowed (specified in the contract).
+    allowed_roles: Vec<String>,
+    /// The result of the transaction. Not allowing view calls here since
+    /// `ViewResultDetails` is not sufficient to verify ACL failure.
+    result: ExecutionFinalResult,
+}
+
+impl TxOutcome {
+    fn assert_success(&self, want: String) {
```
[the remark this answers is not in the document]
mooori: Done in 5eca45c.

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+        log!("{} set_status with message {}", account_id, message);
+        self.records.insert(account_id, message);
+    }
+
+    pub fn get_status(&self, account_id: AccountId) -> Option<String> {
+        log!("get_status for account_id {}", account_id);
+        self.records.get(&account_id).cloned()
+    }
+
+    #[access_control_any(roles(Role::LevelA, Role::LevelC))]
```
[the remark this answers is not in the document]
mooori: Using the `darling` crate key-value pairs of arguments are parsed into a struct:

```rust
// Assume those are the expected arguments for the `my_attr` attribute macro.
#[derive(Debug, FromMeta)]
pub struct MacroArgs {
    arg_1: u8,
    arg_2: u8,
}

// Usage looks like:
#[my_attr(arg_1 = 42, arg_2 = 24)]
fn do_stuff() {
    println!("Hello");
}
```

In the case here, omitting `roles` and writing `#[access_control_any(Role::LevelA, Role::LevelC)]` makes `darling` complain about:

- One missing field: `roles`
- Two unknown fields: `Role::LevelA`, `Role::LevelB`

It _might_ be possible to omit `roles` by either:

- Digging deep into `darling`, `syn`, ...
- Or doing our own argument parsing and error handling/reporting (that's probably not recommended)

---

Changing the key from `roles` to `some_other_value` would be easy.

near-plugins-derive/src/access_control_role.rs
```diff
@@ -0,0 +1,214 @@
...
+const DEFAULT_BOUNDCHECKER_TYPE_NAME: &str = "__AclBoundchecker";
+
+pub fn derive_access_control_role(input: TokenStream) -> TokenStream {
+    // This derive doesn't take attributes, so no need to use `darling`.
+    let input: ItemEnum = parse_macro_input!(input);
+    let ItemEnum {
+        ident, variants, ..
+    } = input;
+
+    let variants = variants.iter().collect::<Vec<_>>();
```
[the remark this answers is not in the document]
mooori: Done in ca5f014.

near-plugins-derive/src/access_controllable.rs
mooori: Done in [f60faa8](https://github.com/aurora-is-near/near-plugins/pull/5/commits/f60faa824d5bb134bded8d9073dd371d065aad5c).

near-plugins-derive/src/access_controllable.rs
mooori: It's a `darling::Error` which doesn't have `to_compile_error()`. [`to_compile_error`](https://docs.rs/syn/1.0.99/syn/struct.Error.html#method.to_compile_error) is a method of `syn::Error`.

The approach using `write_error()` is [recommended](https://github.com/TedDriggs/darling#macro-code).

near-plugins-derive/src/access_controllable.rs
mooori: Good point, added this check in 790824f. There's even [`is_power_of_two()`](https://doc.rust-lang.org/stable/std/primitive.u128.html#method.is_power_of_two).

near-plugins-derive/src/access_controllable.rs
mooori: Done in 1c3a05a.

near-plugins-derive/src/access_controllable.rs
mooori: If the field is not injected we (or the developers) have to manually (de)serialize and read/write the object from/to storage. Probably using [`storage_read`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.storage_read.html) and [`storage_write`](https://docs.rs/near-sdk/latest/near_sdk/env/fn.storage_write.html). We discussed offline to inject the field, to avoid the steps mentioned above.

---

I have not yet checked if it's feasible, but another approach might be:

- Ask the developer to have a field `__acl: ()`. The type this field should ultimately have is defined only inside the attribute macro. So I think developers can't write `__acl: AclType` in their contract.
- Inside the macro, we change the type of this field (instead of injecting a new field).

Probably it wouldn't be a huge win, but makes it more explicit.

mooori: [`near-floating-state`](https://github.com/mfornet/near-floating-state) fits nicely here! Let's stay with injection for now to finalize this PR soon? Then once `near-floating-state` is on crates.io we can introduce it for Acl in a follow-up PR.

near-plugins-derive/src/utils.rs
```diff
@@ -1,18 +1,58 @@
...
+    };
+    if condition_1 {
+        return true;
+    }
+
+    // For a struct `Contract`, `#[near-bindgen]` [generates] `ContractExt`. If
+    // `item` is in an implementation of `ContractExt`, the span number of its
+    // signature differs from the span number of the self token.
+    //
+    // [generates]: https://github.com/near/near-sdk-rs/pull/742
```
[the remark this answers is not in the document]
mooori: An idea for a less hacky approach is asking `near-sdk` to _not_ forward some attributes. It could look like:

```rust
#[near_bindgen(dont_forward(access_control_any))]
impl Contract {
    // ...
}
```

near-plugins/tests/contracts/access_controllable/src/lib.rs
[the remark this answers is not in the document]
mfornet: Noted! Let's keep it simple.

near-plugins/tests/contracts/access_controllable/src/lib.rs
mfornet: Let's keep it simple for now, as it is after commit LGTM.

near-plugins-derive/src/utils.rs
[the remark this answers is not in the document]
mfornet: I like this approach. Shared on near-sdk-rs: https://github.com/near/near-sdk-rs/issues/722#issuecomment-1267651275

mfornet (approved):
LGTM. Merging now, and admin methods will be added in a follow-up PR.

mrLSD (approved):

PR #6, opened 2022-09-16
Fix compilation of integration tests

mooori:
Without these changes integration tests in `near-plugins/tests` don't compile (in my setup).

### DIFF
--- a/near-plugins/src/test_utils.rs
+++ b/near-plugins/src/test_utils.rs
@@ -17,7 +17,7 @@ pub(crate) fn get_context() -> VMContext {
         account_locked_balance: 0,
         storage_usage: 10_000,
         attached_deposit: 0,
-        prepaid_gas: Gas(10u64.pow(18)),
+        prepaid_gas: Gas(1_000_000 * Gas::ONE_TERA.0),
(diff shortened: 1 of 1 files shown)


near-plugins/src/test_utils.rs
```diff
@@ -15,8 +17,8 @@ pub(crate) fn get_context() -> VMContext {
         account_locked_balance: 0,
         storage_usage: 10_000,
         attached_deposit: 0,
-        prepaid_gas: 10u64.pow(18),
-        random_seed: vec![0, 1, 2],
+        prepaid_gas: Gas(10u64.pow(18)),
```
sept-en, line 20: It's better to already use a new available constant. However, it seems we are using a super big value for that:)

```suggestion
        prepaid_gas: 1_000_000 * Gas::ONE_TERRA,
```

sept-en (approved):

near-plugins/src/test_utils.rs
mooori: Done in 114f74d.

mrLSD (approved):

PR #8, opened 2022-10-05
Add functions to manage Acl (super-)admins

mooori:
This PR builds on top of #5 and adds Acl functions related to (super-)admin permissions. A super-admin may act as admin for every role, [these tests](https://github.com/mooori/near-plugins/blob/9bbc751f2bb9ff265aa8f37472b282814463b8fb/near-plugins/tests/access_controllable.rs#L251-L332) provide an overview of super-admin permissions.

## Overview
- Trait methods are added to `AccessControllable`.
- These new methods are implement for a struct with `#[access_controllable]`.
- New methods are tested in `near-plugins/tests/access_controllable.rs`

```
# Command to execute the tests:
cargo test --test access_controllable

# This makes `workspaces` write more than 1G to /tmp (on my machine).
# Depending on the size of your /tmp, a cleanup might be required afterwards.
du --summarize --total --human-readable /tmp/sandbox-*
rm -r /tmp/sandbox-*
```

## Notes on super-admin
Currently only `acl_{add, revoke}_super_admin_unchecked` are added, i.e. only the contract itself may add and revoke super-admins. Potential further additions could be:

```
Allow super-admin to add/revoke other accounts as super-admin.
- acl_add_super_admin(account_id)
- acl_revoke_super_admin(account_id)

Allow a super-admin to renounce their super-admin permission.
- acl_renounce_super_admin()

All of the above are basically just wrappers for
acl_{add, revoke}_super_admin_unchecked which verify the permission of the
caller.
```

## Next up
A follow-up PR to add some missing functions related to roles:

```
acl_grant_role
acl_revoke_role[_unchecked]
acl_renounce_role
```

### DIFF
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -107,6 +107,60 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 }
             }
 
+            fn add_super_admin_unchecked(&mut self, account_id: &::near_sdk::AccountId) -> bool {
+                let flag = <#bitflags_type>::from_bits(<#role_type>::acl_super_admin_permission())
+                    .expect(#ERR_PARSE_BITFLAG);
+                let mut permissions = self.get_or_init_permissions(account_id);
+
+                let is_new_super_admin = !permissions.contains(flag);
+                if is_new_super_admin {
+                    permissions.insert(flag);
+                    self.permissions.insert(account_id, &permissions);
+                    self.add_bearer(flag, account_id);
+
+                    let event = ::#cratename::access_controllable::events::SuperAdminAdded {
+                        account: account_id.clone(),
+                        by: ::near_sdk::env::predecessor_account_id(),
+                    };
+                    event.emit();
+                }
+
+                is_new_super_admin
+            }
+
+            fn is_super_admin(&self, account_id: &::near_sdk::AccountId) -> bool {
+                let permissions = {
+                    match self.permissions.get(account_id) {
+                        Some(permissions) => permissions,
+                        None => return false,
+                    }
+                };
+                let super_admin = <#bitflags_type>::from_bits(<#role_type>::acl_super_admin_permission())
+                    .expect(#ERR_PARSE_BITFLAG);
+                permissions.contains(super_admin)
+            }
+
+            fn revoke_super_admin_unchecked(&mut self, account_id: &::near_sdk::AccountId) -> bool {
+                let flag = <#bitflags_type>::from_bits(<#role_type>::acl_super_admin_permission())
+                    .expect(#ERR_PARSE_BITFLAG);
+                let mut permissions = self.get_or_init_permissions(account_id);
+
+                let was_super_admin = permissions.contains(flag);
+                if was_super_admin {
+                    permissions.remove(flag);
+                    self.permissions.insert(account_id, &permissions);
+                    self.remove_bearer(flag, account_id);
+
+                    let event = ::#cratename::access_controllable::events::SuperAdminRevoked {
+                        account: account_id.clone(),
+                        by: ::near_sdk::env::predecessor_account_id(),
+                    };
+                    event.emit();
+                }
+
+                was_super_admin
+            }
+
             fn add_admin(&mut self, role: #role_type, account_id: &::near_sdk::AccountId) -> Option<bool> {
                 if !self.is_admin(role, &::near_sdk::env::predecessor_account_id()) {
                     return None;
@@ -285,6 +339,20 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
         impl AccessControllable for #ident {
             fn acl_storage_prefix() -> &'static [u8] {
                 (#storage_prefix).as_bytes()
+            }
+
+            #[private]
+            fn acl_add_super_admin_unchecked(&mut self, account_id: ::near_sdk::AccountId) -> bool {
+                self.#acl_field.add_super_admin_unchecked(&account_id)
+            }
+
+            fn acl_is_super_admin(&self, account_id: ::near_sdk::AccountId) -> bool {
+                self.#acl_field.is_super_admin(&account_id)
+            }
+
+            #[private]
+            fn acl_revoke_super_admin_unchecked(&mut self, account_id: ::near_sdk::AccountId) -> bool {
+                self.#acl_field.revoke_super_admin_unchecked(&account_id)
             }
 
             fn acl_add_admin(&mut self, role: String, account_id: ::near_sdk::AccountId) -> Option<bool> {
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -107,6 +107,82 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 }
             }
 
+            fn add_admin(&mut self, role: #role_type, account_id: &::near_sdk::AccountId) -> Option<bool> {
+                if !self.is_admin(role, &::near_sdk::env::predecessor_account_id()) {
+                    return None;
+                }
+                Some(self.add_admin_unchecked(role, account_id))
+            }
+
+            fn add_admin_unchecked(&mut self, role: #role_type, account_id: &::near_sdk::AccountId) -> bool {
+                let flag = <#bitflags_type>::from_bits(role.acl_admin_permission())
+                    .expect(#ERR_PARSE_BITFLAG);
+                let mut permissions = self.get_or_init_permissions(account_id);
+
+                let is_new_admin = !permissions.contains(flag);
+                if is_new_admin {
+                    permissions.insert(flag);
+                    self.permissions.insert(account_id, &permissions);
+                    self.add_bearer(flag, account_id);
+
+                    let event = ::#cratename::access_controllable::events::AdminAdded {
+                        role: role.into(),
+                        account: account_id.clone(),
+                        by: ::near_sdk::env::predecessor_account_id(),
+                    };
+                    event.emit();
+                }
+
+                is_new_admin
+            }
+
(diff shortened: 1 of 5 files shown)


mfornet (changes requested):
Looks good in general. I left a comment about removing unchecked methods from the public contract interface.

mfornet:
FYI: project doesn't compile for me when doing cargo test

mooori:
I'm in the process of removing the `*_unchecked` methods and noticed that:

### Tests rely on calling `*_unchecked` methods

They have been since the first Acl PR. An easy fix is manually exporting `*_unchecked` methods in the test contract.

```rust
// near-plugins/tests/contracts/access_controllable/src/lib.rs

// Helpers to let integration tests call *_unchecked methods.
#[near_bindgen]
impl StatusMessage {
    #[private]
    pub fn acl_add_admin_unchecked(&mut self, role: Role, account_id: AccountId) -> bool {
        self.__acl.add_admin_unchecked(role, &account_id)
    }

    // same for the other *_unchecked methods
}
```

### Developers can no longer use `*_unchecked` for initialization

Hence they would have to call internal methods, for instance:

```rust
#[near_bindgen]
impl Contract {
    #[init]
    pub fn new() -> Self {
        self.__acl.add_super_admin_unchecked(&env::currenct_account_id());
        // ...
    }
}
```

To avoid that, a workaround could be having some method like this in the trait:

```rust
/// Makes the contract a super-admin. It is #[private] in the implementation
/// provided by this trait, i.e. only the contract itself may call this method.
fn acl_make_self_super_admin();
```

Once the contract is super-admin, it can call all _checked_ methods like `acl_add_admin` and `acl_grant_role`.

---

Side note: OpenZeppelin Acl has unchecked methods, e.g. [`_grant_role()`](https://docs.openzeppelin.com/contracts/4.x/api/access#AccessControl-_grantRole-bytes32-address-).

near-plugins/src/access_controllable.rs
```diff
@@ -19,8 +19,73 @@ pub trait AccessControllable {
...
     fn acl_storage_prefix() -> &'static [u8];
 
+    /// Makes `account_id` a super-admin __without__ checking any permissions.
+    /// It returns whether `account_id` is a new super-admin.
+    ///
+    /// Note that there may be zero or more super-admins.
+    ///
+    /// This method is `#[private]` in the implementation provided by this
+    /// crate.
+    fn acl_add_super_admin_unchecked(&mut self, account_id: AccountId) -> bool;
```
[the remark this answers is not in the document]
mooori: Removed in [c5e0320](https://github.com/aurora-is-near/near-plugins/pull/8/commits/c5e0320943e3446c16b2763baff705badc447bdd).

near-plugins-derive/src/access_controllable.rs
```diff
@@ -198,6 +341,52 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 (#storage_prefix).as_bytes()
             }
 
+            #[private]
+            fn acl_add_super_admin_unchecked(&mut self, account_id: ::near_sdk::AccountId) -> bool {
+                self.#acl_field.add_super_admin_unchecked(&account_id)
+            }
```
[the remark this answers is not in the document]
mooori: Removed in [c5e0320](https://github.com/aurora-is-near/near-plugins/pull/8/commits/c5e0320943e3446c16b2763baff705badc447bdd).

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -21,6 +23,26 @@ pub struct StatusMessage {
...
+    //
+    // Once an account is (super-)admin, it may add other admins and grant
+    // roles.
+    //
+    // If needed, It's also possible to grant a role without checks:
+    //
+    // ```
+    // self.__acl.grant_role_unchecked(role, account_id);
+    // ```
+
```
mooori, line 45: Due to removing `acl_*_unchecked` methods from the trait.

I'm not really comfortable with asking contract developers to call `__acl.*_unchecked` to add initial (super-)admins. Instead, we could add some method like the following to the trait `AccessControllable`:

```rust
/// Makes the contract a super-admin.
/// 
/// It is #[private] in the implementation provided by this trait, i.e. only the
/// contract itself may call this method.
fn acl_make_self_super_admin();
```

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -21,6 +23,26 @@ pub struct StatusMessage {
...
+    // from within methods with attribute `#[init]` or `#[private]`.
+    //
+    // Once an account is (super-)admin, it may add other admins and grant
+    // roles.
+    //
+    // If needed, It's also possible to grant a role without checks:
+    //
+    // ```
+    // self.__acl.grant_role_unchecked(role, account_id);
+    // ```
```
mooori, line 44: Due to removing `acl_*_unchecked` methods from the trait.

I'm not really comfortable with asking contract developers to call `self.__acl.*_unchecked` to add initial (super-)admins. Instead, we could add some method like the following to the trait AccessControllable:

```rust
/// Makes the contract a super-admin.
/// 
/// It is #[private] in the implementation provided by this trait, i.e. only the
/// contract itself may call this method.
fn acl_make_self_super_admin();
```

mfornet:
IMO, unchecked methods are an internal detail about our default implementation and not mandatory for every other implementation out there. It is conceivable that other implementations use different mechanisms internally. Arguably unchecked-like methods will be present one way or another, but we should try to keep the main interface as simple as possible. 

> Side note: OpenZeppelin Acl has unchecked methods, e.g. [`_grant_role()`](https://docs.openzeppelin.com/contracts/4.x/api/access#AccessControl-_grantRole-bytes32-address-).

This is a great example of what I try to convey. `_grantRole` is only part of the default implementation provided by OpenZeppelin, but it is not callable or present in the global [Access Control Interface (IAccessControl)](https://docs.openzeppelin.com/contracts/4.x/api/access#IAccessControl). 

> Tests rely on calling *_unchecked methods

We can keep this method behind a testing feature flag.

> Developers can no longer use *_unchecked for initialization
> Proposal: fn acl_make_self_super_admin();

I like the new proposal, even we can take it further and have the method:

```rust
fn acl_init_super_admin(accountId: AccountId);
```

This method can only be called once and will add the account passed as an argument as super admin. In particular, it is mandatory (or strongly recommended) to developers that this is called during initialization.

near-plugins/tests/contracts/access_controllable/src/lib.rs
mfornet: Answered this comment in a comment [here](https://github.com/aurora-is-near/near-plugins/pull/8#issuecomment-1272002777).

mfornet:
> Hm on my machine it compiles and all tests are passing. I'm having near-sdk version 4.0.0 in `Cargo.lock`, is it the same for you? There might be issues when locally 4.0.0.-pre* is installed.
> 
> If it's not related to the version of near-sdk, could you please post more infos about the compilation error?

Check [compilation-error-logs](https://gist.github.com/mfornet/aec3bb9d20de76653e7e70874f29540e/edit) and [Cargo.lock](https://gist.github.com/mfornet/5420b351ed9e01bb2cc7e5bcffb4bbec).

near-plugins/tests/contracts/access_controllable/src/lib.rs
```diff
@@ -49,3 +71,32 @@ impl StatusMessage {
...
+    #[private]
+    pub fn acl_revoke_admin_unchecked(&mut self, role: Role, account_id: AccountId) -> bool {
+        self.__acl.revoke_admin_unchecked(role, &account_id)
+    }
+
+    #[private]
+    pub fn acl_grant_role_unchecked(&mut self, role: Role, account_id: AccountId) -> bool {
+        self.__acl.grant_role_unchecked(role, &account_id)
+    }
+}
```
mfornet, line 102: Let's add this behind [integration-test] feature flag, similar as it is used in the engine today.

mooori:
> ```rust
> fn acl_init_super_admin(accountId: AccountId);
> ```
> 
> This method can only be called once and will add the account passed as an argument as super admin.

Do we want it to be callable _strictly_ once? I think that would require a new field like `initiated: bool` on the Acl struct.

Alternatively we might relax it a bit and say `acl_init_super_admin` succeeds only if there are no super-admins. This condition is met during initialization and additionally when a contract has (by mistake) removed all super-admins. So it's kind of a recovery mechanism.

near-plugins/tests/contracts/access_controllable/src/lib.rs
mooori: https://github.com/aurora-is-near/near-plugins/blob/b2e1d8369059bee4b0e31cb92b965a59a8a2717f/near-plugins/tests/access_controllable.rs#L25

The contract containing these methods is compiled and deployed via [`workspaces::compile_project`](https://docs.rs/workspaces/latest/workspaces/fn.compile_project.html) which has only one parameter: `project_path: &str`. AFAIK this prevents us from working with features in this contract, since we'd have to enable required features by default (see below).

The following works, though if the `integration-test` feature needs to be enabled by default, I'm not sure if it makes sense to have it?

```
[features]
default = ["integration-test"]
integration-test = []
```

---

It might be worth a feature request for `compile_project` to have an additional parameter `&[&str]` that specifies the features to be enabled.

mooori:
> Regarding the compilation errors: After removing `Cargo.lock` in addition to `cargo clean` I could reproduce the errors. Skimming over them, looks like they're all in dependencies. I'm looking into it...

Running `cargo test` in a fresh directory pulled in different, inconsistent versions of `near*` dependencies. Updating the `workspaces` dev-dependency to the latest version should fix it (done in the latest commit).

mfornet (approved):
LGTM. I'll revisit later about having init_super_admin as a public method, hidden under a private modifier, I think this is no needed and can be internal. But let's move forward with this PR for now.

mrLSD (approved):

PR #15, opened 2022-10-11
Add CI

### LINKED ISSUE #10: CI: build and test contracts
It would be good to have all the contract builds & tests running in CI for `near-plugins`. This is especially mandatory as this library will be used in other 3rd party contracts.

mfornet:
Use GitHub workflows for local CI. 

This PR is targeting `acl` since there is a dependency problem on master that is solved in that branch.

Closes #10

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,50 @@
+name: Rust Contracts
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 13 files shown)


mfornet:
@mooori good suggestion. Already changed in settings to force this behavior. PRs will use commit squashing.

mrLSD (approved):

PR #23, opened 2022-10-18
Make clippy check tests and fix warnings

mooori:
Adding `--all-targets` makes clippy check tests. This results in a [quite a few warnings](https://github.com/aurora-is-near/near-plugins/actions/runs/3272111224/jobs/5382724238) as previously tests were ignored.

TODO: fix warnings


mooori:
I'm having trouble pushing to that branch to trigger a CI rerun. Closing this in favor of #64.

PR #24, opened 2022-10-18
Re-export crate to avoid dependency in contract

mooori:


near-plugins/tests/contracts/access_controllable/Cargo.toml
```diff
@@ -7,7 +7,6 @@ edition = "2018"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-bitflags = "1.3"
```
mooori, line 10: We no longer need to ask contract developers to do this.

mfornet (approved):

mrLSD (approved):

PR #26, opened 2022-10-18
Use double underscore in generated variables

### LINKED ISSUE #17: Variables collision with macro
The double underscore is used [here](https://github.com/aurora-is-near/near-plugins/blob/d822c2f051ccdc03b50f206650524daf43ae7526/near-plugins-derive/src/ownable.rs#L111) to avoid a collision with local variables. But [here](https://github.com/aurora-is-near/near-plugins/blob/d822c2f051ccdc03b50f206650524daf43ae7526/near-plugins-derive/src/pausable.rs#L159) the double underscore is missed, so it is possible to write code like that:
```
  #[pause]
  fn increase_1(&mut self) {
      check_paused = false;
      println!("check_paused: {}", check_paused);
      self.counter += 1;
  }
```

mfornet:
Use __check_paused to avoid collision with local variables.

Closes #17

### DIFF
--- a/near-plugins-derive/src/pausable.rs
+++ b/near-plugins-derive/src/pausable.rs
@@ -126,9 +126,9 @@ pub fn pause(attrs: TokenStream, item: TokenStream) -> TokenStream {
     let bypass_condition = get_bypass_condition(&args.except);
 
     let check_pause = quote!(
-        let mut check_paused = true;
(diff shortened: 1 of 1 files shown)


sept-en (approved):
LGTM!

mooori (approved):

mrLSD (approved):

PR #37, opened 2022-11-04
Migrate from `near_sdk::collections` to `near_sdk::store`

mooori:
This only affects the Acl plugin, other plugins aren't using `near_sdk::collections`.

## Motivation
`near_sdk::collections` [will be put under the legacy feature flag](https://github.com/near/near-sdk-rs/pull/923) and its successor is `near_sdk::store`. Once a contract uses a plugin based on `collections`, it will require efforts to migrate that contract to a new plugin version based on `store`. To avoid that, let's migrate `near-plugins` itself to `near_sdk::store` before any plugins are used in production.

## `near_sdk::store`
Main differences to `near_sdk::collections` are described in the last section of [this post](https://github.com/near/near-sdk-rs/discussions/829).

## Feature flag `unstable`
To use `store` it's currently still required to enable the `unstable` feature for `near_sdk`. Seems like `store` will be stabilized in sdk version 4.1 (currently it's at 4.0.0).

### Usage patterns are more idiomatic for rust
For instance maps now have `get_mut` which allows removing patterns like `get element -> modify it -> write back to map`.

### Clones
Using `store` without changing any function signatures requires cloning values at various places since some `near-sdk` functions previously handling references now handle owned values and vice versa. For instance `collections::UnorderedSet::insert` has parameter `element: &T` while `store::UnorderedSet::insert` has parameter `element: T`.

A follow-up PR could change some signatures of internal functions of the Acl plugin to avoid some of these clones.

## Review
It might be more convenient to look at the commits individually.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,8 @@ keywords = ["near", "smart contract", "plugin"]
 
 [workspace.dependencies]
 bitflags = "1.3"
-near-sdk = "4.0.0"
+# Feature `unstable` is required to use `near_sdk::store`.
+near-sdk = { version = "4.0.0", features = ["unstable"] }
 near-plugins-derive = { path = "near-plugins-derive" }
 serde = "1"
 anyhow = "1.0"
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -55,7 +55,7 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
         #[derive(::near_sdk::borsh::BorshDeserialize, ::near_sdk::borsh::BorshSerialize)]
         struct #acl_type {
             /// Stores permissions per account.
-            permissions: ::near_sdk::collections::UnorderedMap<
+            permissions: ::near_sdk::store::UnorderedMap<
                 ::near_sdk::AccountId,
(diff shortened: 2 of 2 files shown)


mfornet (approved):
It is unfortunate the introduction of new clones. Mostly because I expect, inside these methods, the object is fully serialized, so it is first cloned, then serialized. This is feedback we can pass on to the near-sdk-rs team.

sept-en (approved):
LGTM! Looks cleaner now. The only thing is to perhaps wait for API to stabilize, as we want to have a minimal impact on future upgrades.

mooori:
I've created #38 for changing internal function signature to get rid of some of the new clones. I should have more insights afterwards and will leave another comment then.

---

We're using `UnorderedSet` and `UnordereMap` from `near_sdk::store`. It seems like in our version of near-sdk (4.0.0) they're unstable just due to outstanding optimizations, testing and docs. They should be stabilized in the upcoming 4.1.1. without changes affecting `near-plugins` ([#922](https://github.com/near/near-sdk-rs/pull/922), [#924](https://github.com/near/near-sdk-rs/pull/924)).

So I assume it's fine to merge this PR now. @sept-en wdyt?

sept-en:
@mooori makes sense to me. This PR is ok to merge for sure.

mrLSD (approved):

PR #42, opened 2022-11-23
Add Acl plugin to the Readme

### LINKED ISSUE #41: Update README with the ACL documentation
The README still contains ACL as a feature to be implemented in the roadmap. As this is already implemented, I think this part should be replaced with the appropriate ACL documentation.

mooori:
An introduction of the Acl plugin was missing - it was still listed unter Roadmap.

Closes #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,87 @@ impl Counter {
 To upgrade the contract first call `up_stage_code` passing the binary as first argument serialized as borsh. Then call `up_deploy_code`.
 This functions must be called from the owner.
 
+### [AccessControllable](/near-plugins/src/access_controllable.rs)
+
+Enables role based access control for contract methods. A method with restricted access can only be called _successfully_ by accounts that have been granted one of the whitelisted roles. If a restricted method is called by an account with insufficient permissions it panics.
+
+Each role is managed by admins who may grant the role to accounts and revoke it from them. In addition, there are super admins that have admin permissions for every role.
+
+The sets of accounts that are (super) admins and grantees are stored in the contract's state.
+
+```rust
+/// Roles are represented by enum variants.
+/// 
+/// Deriving `AccessControlRole` ensures `Role` can be used in
+/// `AccessControllable`.
+#[derive(AccessControlRole, Deserialize, Serialize, Copy, Clone)]
+#[serde(crate = "near_sdk::serde")]
+pub enum Role {
+    SkipperByOne,
+    SkipperByAny,
+    Resetter,
+}
+
+#[access_control(role_type(Role))]
+#[near_bindgen]
+#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+pub struct Contract {
+    counter: u64,
+}
+
+#[near_bindgen]
+impl Counter {
+    /// Setup access control in the constructor.
+    #[init]
+    pub fn new() -> Self {
+        let contract = Self {
+            counter: 0,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -178,6 +178,87 @@ impl Counter {
 To upgrade the contract first call `up_stage_code` passing the binary as first argument serialized as borsh. Then call `up_deploy_code`.
 This functions must be called from the owner.
 
+### [AccessControllable](/near-plugins/src/access_controllable.rs)
+
+Enables role based access control for contract methods. A method with restricted access can only be called _successfully_ by accounts that have been granted one of the whitelisted roles. If a restricted method is called by an account with insufficient permissions it panics.
```
sept-en, line 183:
```suggestion
Enables role-based access control for contract methods. A method with restricted access can only be called _successfully_ by accounts that have been granted one of the whitelisted roles. If a restricted method is called by an account with insufficient permissions, it panics.
```

README.md
```diff
@@ -178,6 +178,87 @@ impl Counter {
...
+        near_sdk::require!(
+            Some(true) == contract.acl_add_admin(Role::SkipperByOne.into(), admin_account_id),
+            "Failed to add admin",
+        );
+
+        // Grant a role. Also possible since the contract is super admin.
+        near_sdk::require!(
+            Some(true) == contract.acl_grant_role(Role::SkpperByAny.into(), grantee_account_id),
+            "Failed to grant role",
+        );
```
sept-en, line 236: I think it's worth adding variables for `admin_account_id` and `grantee_account_id` just to show that these are user-defined.

E.g.:
```rust
// Define the account you would like to be an admin
let skipper_by_one_admin_account_id: AccountId = "alice.near".parse().unwrap();
```

It's also worth adding a comment for the "add an admin" part that we are adding an admin just for the specific role. Even though it's mentioned before that admins and super-admins are different in the ACL hierarchy, I don't think the comment would be redundant.

README.md
```diff
@@ -178,6 +178,87 @@ impl Counter {
...
+    }
+
+    /// This method has no access control. Anyone can call it successfully.
+    pub fn increase(&mut self) {
+       self.counter += 1; 
+    }
+
+    /// Only an account which was granted any of the whitelisted roles may
+    /// successfully call this method.
```
sept-en, line 245: There's also a third role that exists, so "any of the whitelisted roles" could be confusing. It's better to explain roles for the example explicitly.
```suggestion
    /// Only an account that was granted either `SkipperByOne` or `SkipperByAny` roles may
    /// successfully call this method.
```

sept-en (changes requested):

mooori:
@sept-en I've marked above comments as resolved since they were adressed in be31e02.

README.md
```diff
@@ -178,6 +178,96 @@ impl Counter {
...
+            __acl: Default::default(),
+        };
+
+        // Add the contract itself as super admin.
+        near_sdk::require!(
+            contract.acl_init_super_admin(near_sdk::env::predecessor_account_id()),
+            "Failed to initialize super admin",
+        );
+
+        // Specify an account to be added as admin for a specifc role.
```
sept-en, line 226: A small typo
```suggestion
        // Specify an account to be added as admin for a specific role.
```

sept-en (approved):

mfornet (approved):

mfornet:
@mooori I missed your comment in #43 and merged it. Can you please fix conflicting files here and merge?

Issue #20, opened 2022-10-17
Upgradeability: add delay on code upgrade


sept-en:
The functionality of staging and deploying the code is implemented by the `Upgradable` part of NEAR plugins. As the code is being staged and deployed in separate functions, it allows for implementing delayed upgrades. So to achieve this, an additional storage variable could be introduced that will contain a value for the delayed upgrade. However, it's worth noting that in this case, changing the value of this variable should also be allowed only via contract upgrade or delayed-value-upgrade (so to change the delay-upgrade period, you need to wait for the same expected delay).

PR #44, opened 2022-11-24
Add support for the delayed upgrade

karim-en:
Implement [#20](https://github.com/aurora-is-near/near-plugins/issues/20)

Added apis:
```
- up_storage_prefix();
- up_get_delay_status(&self) -> UpgradableDurationStatus;
- up_init_staging_duration(staging_duration: near_sdk::Duration);
- up_stage_update_staging_duration(staging_duration: near_sdk::Duration);
- up_apply_update_staging_duration();

```

Added storage keys:
```
- "StagingDuration":  to save the delay duration of deploying staged code.
- "UpdateStagingDuration": to save the staged delay duration update.
- "StagingTimestamp": to save the allowed timestamp to apply the staged duration or code.
```


near-plugins-derive/src/upgradable.rs
```diff
@@ -21,19 +22,29 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
                 (#code_storage_key).as_bytes().to_vec()
             }
 
+            fn up_allowed_timestamp_storage_key(&self) -> Vec<u8> {
+                (#allowed_timestamp_storage_key).as_bytes().to_vec()
+            }
+
             #[#cratename::only(owner)]
-            fn up_stage_code(&mut self, #[serializer(borsh)] code: Vec<u8>) {
+            fn up_stage_code(&mut self, #[serializer(borsh)] code: Vec<u8>,  #[serializer(borsh)] delay_timestamp: u64) {
```
mfornet, line 41: Orthogonal to this PR. I wonder if we should avoid borsh here and use json, passing the code in base64. It will be more inefficient, but better for readability.

mfornet, line 41: The staging time should be independent from the submission of the code. Otherwise, this protection is meaningless. Changing staging time should require higher privileges than staging code. This way users can trust that code won't change without having a notice period.

See a similar proposal with this feature here: https://github.com/near/NEPs/pull/123

mfornet (changes requested):

birchmd (approved):
LGTM

near-plugins-derive/src/upgradable.rs
```diff
@@ -21,20 +25,56 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+            }
+
+            fn up_update_staging_duration_storage_key(&self) -> Vec<u8> {
+                (#update_staging_duration_storage_key).as_bytes().to_vec()
+            }
+
+            fn up_update_staging_duration_timestamp_storage_key(&self) -> Vec<u8> {
+                (#update_staging_duration_timestamp_storage_key).as_bytes().to_vec()
+            }
+
```
mooori, line 66: Could these values be collected in a struct? Maybe something like:

```rust
#[derive(BorshSerialize, BorshDeserialize)]
struct Delay {
    /// After staging code at least this amount of time must pass before the
    /// code can be deployed.
    delay: Option<Duration>,
    /// If code is staged, it cannot be deployed before this date.
    earliest_deployment: Option<near_sdk::Timestamp>,
}

// Borsh methods for (de)serialization from/to storage:
// * https://docs.rs/borsh/0.9.3/borsh/ser/trait.BorshSerialize.html#method.try_to_vec
// * https://docs.rs/borsh/0.9.3/borsh/de/trait.BorshDeserialize.html#method.try_from_slice
```

Possible benefits:
- Reduce the number of methods in the `Upgradable` trait: there's only one storage key and one getter to get the whole struct might be enough.
- Less (de)serialization in the implementation of `Upgradable`, which makes it easier to follow what's happening.

near-plugins-derive/src/upgradable.rs
```diff
@@ -49,9 +89,84 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+                if staging_timestamp < near_sdk::env::block_timestamp() {
+                    near_sdk::env::panic_str(
+                        format!(
+                            "Upgradable: Update duration too early: staging ends on {}",
+                            staging_timestamp
+                        )
+                        .as_str(),
+                    );
+                }
```
mooori, line 149: I feel like having a separate duration for updating the staging duration adds a lot of complexity. That's not mentioned in the NEP123, so just wondering if it's needed?

If the delay duration for deployments should be changeable, maybe it can be simplified by not letting the update affect a staged code upgrade?

mooori (commented):
I think the [README](https://github.com/aurora-is-near/near-plugins/blob/c4db1654ffeb8736c009d967c5486523306c265a/README.md) should be updated to explain how to use `Upgradable` with delay. If I’m not mistaken, it can not be used without delay anymore (`up_deploy_code` panics if there’s no staging timestamp).

near-plugins-derive/src/upgradable.rs
karim-en: This is a double-edged sword, logically those values aren't used together so extra serialization will be done, but on the other hand, I agree that the code will look more elegant.

@birchmd Based on your experience of managing the storage in the Aurora engine, do you think that grouping keys into structures is a good idea? Were there any problems with migration due to this?

birchmd: The engine keeps values under separate keys for the most part (as opposed to multiple values in a struct); in fact there are some examples where it probably would have been better to group the values under a single key because they are logically correlated (for example account nonce and balance are almost always both needed together).

In general I think it is more important for the storage layout to match the logical usage of the data (data that is accessed together is stored together), as opposed to grouping everything in one value for code reasons. That said, maybe the code could be made to look a little nicer while keeping the values under separate keys. In the engine we use an enum to separate out the different storage keys we have, but then only a single method for accessing the storage (this method takes the enum as an argument). You can also have specialized read and write methods for common types to hide the serialization logic at the main call-site. In this case this might look something like

```rust
enum UpgradableStorageKey {
    StagingTimestamp,
    StagingDuration,
    UpdateStagingTimestamp,
    UpdateStagingDuration,
}

impl AsRef<[u8]> for UpgradableStorageKey { ... }

fn up_get_update_staging_duration(&self) -> Option<near_sdk::Duration> {
    read_duration(UpgradableStorageKey::UpdateStagingDuration)
}

fn up_get_update_staging_duration_timestamp(&self) -> Option<near_sdk::Timestamp> {
    read_timestamp(&UpgradableStorageKey::UpdateStagingTimestamp)
}

fn read_duration(key: &UpgradableStorageKey) -> Option<near_sdk::Duration> {
    near_sdk::env::storage_read(key.as_ref()).map(...)
}
```

What would you think about that @mooori ?

mooori: That’s a neat way to hide interaction with storage while at the same time avoiding extra de-serialization!

Since near-plugins already depends on near-sdk, we could derive [BorshStorageKey](https://docs.rs/near-sdk/latest/near_sdk/derive.BorshStorageKey.html) for the enum.

Either way, we should be careful not to use storage keys that conflict which storage keys defined by the contract in which the plugin is included. Here’s how we did that in the Acl plugin:

https://github.com/aurora-is-near/near-plugins/blob/7454bbcfcdb50addee5fefe3742879d1fac75c81/near-plugins-derive/src/access_controllable.rs#L83-L99

karim-en: @mooori I've refactored the storage management for the upgradable plugin here a0d0e494f2203ffacee7bd422187d3b6cca96322

mooori (commented):
How about introducing helper functions for storage writes too? It’s a nit, but I think it would improve readability.

```rust
#[inline]
fn up_set_duration(key: __UpgradableStorageKey, duration: &near_sdk::Duration) {
  near_sdk::env::storage_write(self.up_storage_key(key).as_ref(), duration.to_be_bytes());
}

// Same for `fn up_set_timestamp`.
```

These functions would be the counterparts of `up_get_{duration, timestamp}`.

---

Please remember to update the _Upgradable_ section in the [root README](https://github.com/aurora-is-near/near-plugins/blob/ab683721fb72626bcd1e0399b922adae435b81db/README.md).

near-plugins/src/upgradable.rs
```diff
@@ -20,29 +20,15 @@
...
-    /// By default b"__DURATION__" is used.
-    fn up_staging_duration_storage_key(&self) -> Vec<u8>;
-
-    /// Key of storage slot to save the staged delay duration update.
-    /// By default b"__UPDATE_DURATION__" is used.
-    fn up_update_staging_duration_storage_key(&self) -> Vec<u8>;
-
-    /// Key of storage slot to save the allowed timestamp to apply the staged duration update.
-    /// By default b"__UPDATE_DURATION_TIMESTAMP__" is used.
-    fn up_update_staging_duration_timestamp_storage_key(&self) -> Vec<u8>;
```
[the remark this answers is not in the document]
mooori: On second thought I'd argue it's fine to remove these functions. Someone who implements the trait manually might work with different internal data structures and use a different storage layout.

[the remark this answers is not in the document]
birchmd: I agree that storage should be an implementation detail, not a core part of the trait.

karim-en:
> How about introducing helper functions for storage writes too? It’s a nit, but I think it would improve readability.
> 
> ```rust
> #[inline]
> fn up_set_duration(key: __UpgradableStorageKey, duration: &near_sdk::Duration) {
>   near_sdk::env::storage_write(self.up_storage_key(key).as_ref(), duration.to_be_bytes());
> }
> 
> // Same for `fn up_set_timestamp`.
> ```
> 
> These functions would be the counterparts of `up_get_{duration, timestamp}`.
> 
> Please remember to update the _Upgradable_ section in the [root README](https://github.com/aurora-is-near/near-plugins/blob/ab683721fb72626bcd1e0399b922adae435b81db/README.md).

fixed in : 
6489041834e6f7040f583b6795f15eebfc4cee30
3a1f09ae8c752f110adc5a8e3db688cc63a18781

near-plugins-derive/src/upgradable.rs
```diff
@@ -7,39 +7,91 @@ use syn::{parse_macro_input, DeriveInput};
...
+
         #[near_bindgen]
         impl Upgradable for #ident {
-            fn up_storage_key(&self) -> Vec<u8>{
-                (#code_storage_key).as_bytes().to_vec()
+            fn up_storage_prefix(&self) -> &'static [u8] {
+                (#storage_prefix).as_bytes()
+            }
+
+            fn up_get_duration_status(&self) -> #cratename::UpgradableDurationStatus {
```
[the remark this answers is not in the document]
karim-en: Fixed in 209bf87388b543f39ee2a584865575575b035f8d

near-plugins-derive/src/upgradable.rs
```diff
@@ -7,39 +7,91 @@ use syn::{parse_macro_input, DeriveInput};
...
+            Code,
+            StagingTimestamp,
+            StagingDuration,
+            UpdateStagingDuration,
+            UpdateStagingDurationTimestamp,
+        }
+
+        impl #ident {
+            fn up_get_timestamp(&self, key: __UpgradableStorageKey) -> Option<::near_sdk::Timestamp> {
+                near_sdk::env::storage_read(self.up_storage_key(key).as_ref()).map(|staging_timestamp_bytes| {
```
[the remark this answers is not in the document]
karim-en: Fixed in 209bf87388b543f39ee2a584865575575b035f8d

near-plugins-derive/src/upgradable.rs
```diff
@@ -7,39 +7,91 @@ use syn::{parse_macro_input, DeriveInput};
...
+                    u64::from_be_bytes(staging_timestamp_bytes.try_into().unwrap_or_else(|_|
+                        near_sdk::env::panic_str("Upgradable: Invalid u64 timestamp format"))
+                    )
+                })
+            }
+
+            fn up_get_duration(&self, key: __UpgradableStorageKey) -> Option<::near_sdk::Duration> {
+                near_sdk::env::storage_read(self.up_storage_key(key).as_ref()).map(|staging_duration_bytes| {
```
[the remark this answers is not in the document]
karim-en: Fixed in 209bf87388b543f39ee2a584865575575b035f8d

near-plugins-derive/src/upgradable.rs
```diff
@@ -49,9 +101,54 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
 
             #[#cratename::only(owner)]
             fn up_deploy_code(&mut self) -> near_sdk::Promise {
+                let staging_timestamp = self.up_get_timestamp(__UpgradableStorageKey::StagingTimestamp).unwrap_or(0);
```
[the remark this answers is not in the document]
karim-en: Fixed in 209bf87388b543f39ee2a584865575575b035f8d

near-plugins-derive/src/upgradable.rs
```diff
@@ -50,9 +121,64 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+            fn up_stage_update_staging_duration(&self, staging_duration: Option<near_sdk::Duration>) {
+                if let Some(staging_duration) = staging_duration {
+                    let current_staging_duration = self.up_get_duration(__UpgradableStorageKey::StagingDuration)
+                        .unwrap_or_else(|| ::near_sdk::env::panic_str("Upgradable: staging duration isn't initialized"));
+
+                    let staging_duration_timestamp = near_sdk::env::block_timestamp() + current_staging_duration;
+                    self.up_set_timestamp(__UpgradableStorageKey::StagingTimestamp, staging_duration_timestamp);
+                    self.up_set_duration(__UpgradableStorageKey::UpdateStagingDuration, staging_duration);
+                } else {
+                    near_sdk::env::storage_remove(&self.up_storage_key(__UpgradableStorageKey::UpdateStagingDuration));
```
birchmd, line 157: If the deletion of the delay key is immediate, couldn't the owner work around the delay period in updating the delay key by calling `up_stage_update_staging_duration(None)` and then `up_init_staging_duration(new_duration)`? This seems like a loop hole we do not want. I think either (a) deleting the key should be subject to the same delay as setting a new value or (b) it should not be allowed to delete the delay key at all (setting a delay of 0 would be sort of like removing the delay functionality).

near-plugins-derive/src/upgradable.rs
```diff
@@ -18,29 +20,98 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+                near_sdk::require!(
+                    self.up_get_duration(__UpgradableStorageKey::UpdateStagingDuration).is_none(),
+                    "Upgradable: an update of staging duration is in progress"
+                );
+
+                let timestamp = near_sdk::env::block_timestamp() + self.up_get_duration(__UpgradableStorageKey::StagingDuration).unwrap_or(0);
+
                 if code.is_empty() {
-                    near_sdk::env::storage_remove(self.up_storage_key().as_ref());
+                    near_sdk::env::storage_remove(self.up_storage_key(__UpgradableStorageKey::Code).as_ref());
```
mooori, line 104: Despite removing staged code a new `StagingTimestamp` is set below. Shouldn't removing staged code also remove the `StagingTimestamp`?

near-plugins-derive/src/upgradable.rs
```diff
@@ -50,9 +121,64 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+            }
+
+            #[#cratename::only(owner)]
+            fn up_stage_update_staging_duration(&self, staging_duration: Option<near_sdk::Duration>) {
+                if let Some(staging_duration) = staging_duration {
+                    let current_staging_duration = self.up_get_duration(__UpgradableStorageKey::StagingDuration)
+                        .unwrap_or_else(|| ::near_sdk::env::panic_str("Upgradable: staging duration isn't initialized"));
+
+                    let staging_duration_timestamp = near_sdk::env::block_timestamp() + current_staging_duration;
+                    self.up_set_timestamp(__UpgradableStorageKey::StagingTimestamp, staging_duration_timestamp);
```
mooori, line 154: When code is staged, this line shifts the earliest time at which the code can be deployed? What's the reason for that?

near-plugins-derive/src/upgradable.rs
```diff
@@ -50,9 +121,64 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+                    self.up_set_duration(__UpgradableStorageKey::UpdateStagingDuration, staging_duration);
+                } else {
+                    near_sdk::env::storage_remove(&self.up_storage_key(__UpgradableStorageKey::UpdateStagingDuration));
+                }
+            }
+
+            #[#cratename::only(owner)]
+            fn up_apply_update_staging_duration(&self) {
+                let staging_timestamp = self.up_get_timestamp(__UpgradableStorageKey::StagingTimestamp)
```
mooori, line 163: `StagingTimestamp` is also used in `up_deploy_code`. Can we use _one_ timestamp to control _both_ deploying new code and setting a new delay? I expected there to be two separate timestamps.

near-plugins-derive/src/upgradable.rs
mooori: Initially I was thinking the same, though seems like `UpdateStagingDuration` holds the value that shall become the new delay. Then removing it would have the effect of aborting a staged update of the delay. If that's the case, I think the name `UpdateStagingDuration` is misleading.

mooori (commented):
How about using the term `delay` instead of `staging duration` in the code? I think that could make things clearer, for instance:

```rust
fn up_stage_new_delay(&self, delay: Option<near_sdk::Duration>);

// instead of

fn up_stage_update_staging_duration(&self, staging_duration: Option<near_sdk::Duration>);
```

---

There are events for `StageCode` and `DeployCode`. 

https://github.com/aurora-is-near/near-plugins/blob/6489041834e6f7040f583b6795f15eebfc4cee30/near-plugins/src/upgradable.rs#L90-L95

https://github.com/aurora-is-near/near-plugins/blob/6489041834e6f7040f583b6795f15eebfc4cee30/near-plugins/src/upgradable.rs#L110-L115

I think it would make sense to also have events like `StageDelay` and `SetDelay`

karim-en:
> How about using the term `delay` instead of `staging duration` in the code? I think that could make things clearer, for instance:
> 
> ```rust
> fn up_stage_new_delay(&self, delay: Option<near_sdk::Duration>);
> 
> // instead of
> 
> fn up_stage_update_staging_duration(&self, staging_duration: Option<near_sdk::Duration>);
> ```
> 
> There are events for `StageCode` and `DeployCode`.
> 
> https://github.com/aurora-is-near/near-plugins/blob/6489041834e6f7040f583b6795f15eebfc4cee30/near-plugins/src/upgradable.rs#L90-L95
> 
> https://github.com/aurora-is-near/near-plugins/blob/6489041834e6f7040f583b6795f15eebfc4cee30/near-plugins/src/upgradable.rs#L110-L115
> 
> I think it would make sense to also have events like `StageDelay` and `SetDelay`

The term `staging duration` is already used in the [NEP](https://github.com/near/NEPs/pull/123) [UPGRADE](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/upgrade/mod.rs), so I am skeptical about adding new terminology

near-plugins-derive/src/upgradable.rs
karim-en: I thought that this change was requested by you and Michael to use just one tuple of `(duration, timestamp)`

near-plugins-derive/src/upgradable.rs
karim-en: The same as for staged code, it will be just an extra operation without any benefits. but it also could be useful to see the last timestamp as historical info.

near-plugins-derive/src/upgradable.rs
```diff
@@ -50,9 +121,64 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
...
+                }
+
                 near_sdk::Promise::new(near_sdk::env::current_account_id())
                     .deploy_contract(self.up_staged_code().unwrap_or_else(|| ::near_sdk::env::panic_str("Upgradable: No staged code")))
             }
+
+            #[#cratename::only(owner)]
+            fn up_init_staging_duration(&self, staging_duration: near_sdk::Duration) {
+                near_sdk::require!(self.up_get_duration(__UpgradableStorageKey::StagingDuration).is_none(), "Upgradable: staging duration was already initialized");
+                self.up_set_staging_duration_unchecked(staging_duration);
```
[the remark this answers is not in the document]
karim-en: The `up_set_staging_duration_unchecked` cloud be used in the implementation of contracts to be able to set the duration manually with a custom scenario or access.

near-plugins-derive/src/upgradable.rs
karim-en: Ok, maybe we could rename it to  `StagedStagingDuration` or `NewStagingDuration`?

near-plugins-derive/src/upgradable.rs
karim-en: This is a side effect of using one `Timestamp` for both actions (stage code and stage new duration).
This can be avoided by reverting the requested changes and returning the second `timestamp`, so it will be possible to stage a new duration and code in parallel.

Also as a solution, we could prevent calling this function if the staged code is not empty, but removing the code from the storage after deployment could take extra gas so there is a risk to get out of gas problem.

near-plugins-derive/src/upgradable.rs
mooori: `NewStagingDuration` sounds good to me. Though before renaming it, maybe let’s make sure we’re all on the same page regarding the delay update mechanism (see comments below).

near-plugins-derive/src/upgradable.rs
mooori: The proposed idea ([described here](https://aurora-labs-group.slack.com/archives/C02818D9R8T/p1671215546478119?thread_ts=1671209521.724339&cid=C02818D9R8T)) was:

- When there’s no code staged, updates of the delay are immediate (i.e. no delay to update the delay).
- Don’t allow updating the delay when code is staged for an upgrade.

Then having just one tuple of `(duration, timestamp)` is due to not delaying updates of the delay.

Using one timestamp for both code and delay seems prone to introducing unintended side effects.

near-plugins-derive/src/upgradable.rs
mooori: > we could prevent calling this function if the staged code is not empty

This was part of the proposed idea, see [this comment](https://github.com/aurora-is-near/near-plugins/pull/44#discussion_r1090401902).

near-plugins-derive/src/upgradable.rs
karim-en: But in this case, the owner can just update the delay and set it to `0` and then call `stage_code` and `deploy_code` in the same transaction. The users who monitor this contract will not have enough time to do some emergency actions eg. withdraw assets

birchmd: Delayed updates are always a double-edged sword. As you have pointed out, it could lead to sudden changes in the contract that users may not want if the owner is malicious. But in the case where a bug fix to some exploit is being deployed the owner being able to take swift action is important. Otherwise the owner will need to leave the contract paused for a long time while the fix is staged and deployed.

karim-en: This case could be covered by using an ACL, by adding some role eg. `UnrestrictedUpgrade`, and assigning this role to some DAO account.

birchmd: As @mooori mentioned in the slack thread linked above, ACL could also be used to make updating the staging duration a separate role from staging/deploying upgrades. If a DAO account was given the `DurationUpdater` role, then this would have the same security properties as your proposed `UnrestrictedUpgrade` (it remains secure so long as the DAO is not compromised), but the API would be simpler because we would not have this "delayed change of the delay" complexity.

We do plan on migrating the upgradable plugin from using `Ownable` to using `ACL` instead, but we'd like to have this PR land first to avoid merge conflicts.

sept-en: @birchmd I agree it's a double-edged sword. I think the initial idea for this delayed upgrade is exactly to provide the capability not to have the option to update faster than the delay. This could be considered a good alternative to the common practice in NEAR to remove all FAK for the accounts to show that it's locked. In this case it could be time-locked (not requiring a protocol upgrade). From my perspective, the only way to do it is to disallow any updates until the time is passed. To prevent exploits from having an impact, as you mentioned, developers may utilize pauseability functionality and, let's say, pause only some features until the upgrade is landed.

Anyway, the idea is that delayed upgrades functionality is not enforced. As a developer, you are able to choose whether you want this feature or not.

birchmd: > it could be time-locked ... disallow any updates until the time is passed.

Interesting, thanks for presenting it this way. This sounds more general than just upgradable. It sounds like a general feature that developers may want to apply to other properties (e.g. changing roles).

Our main concern with the current design is the complexity in having the basic upgradable trait having two pairs of functions (stage+deploy of code and stage+deploy of staging duration). If we were to factor out this "time locked" feature as a separate plugin then maybe both would be individually easier to parse and the composition of them would maybe also be easy to parse.

On the other hand, maybe writing a general "time locked" plugin would be too challenging and would just lead us down a deeper design rabbit hole. So, nevermind.

Something else has occurred to me: I'm not sure that the current implementation quite gives the guarantee we want. Suppose the staging duration is equal to `D` and the owner stages and update to change it to `D / 3` at time `T`. We know that the new `D / 3` duration does not take effect until time `T + D`, but what if the owner also stages new code to be deployed at time `T + 2D/3`? Then at time `T + D` would it be allowed for the owner to make a batch transaction where the new duration and the new code go live at the same time? (Since the code was staged at `T + 2D/3` and the deploy would be called at `T + D`, so the check `T + D - (T + 2D/3) = D / 3 >= D / 3` (the new duration) would pass.)

In that example, then the new duration sort of takes effect immediately in the sense that new code only needs to be staged for the new duration before it can be deployed.

This could be addressed by not allowing code to be staged while a new duration is staged. And as we have discussed, the reverse should also be true (no duration change should be allowed to be staged while code is staged).

Sorry for the long comment. TLDR: If you feel strongly that having this "time lock" property is important then we should keep it, and we should make staging duration and code changes mutually exclusive.

birchmd: For what it's worth, openzeppelin actually does have a general `TimelockController`, so maybe it's not a completely ridiculous idea to pursue that. But I think we should worry about that separately and just try to get this PR merged for the time being.

https://docs.openzeppelin.com/contracts/4.x/access-control#delayed_operation

karim-en: If we compare our implementation with the one from [nep](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/upgrade/mod.rs), you will notice that the validation on deploying the code is different.
[Here in near-sdk](https://github.com/near/near-sdk-rs/blob/master/near-contract-standards/src/upgrade/mod.rs#L73) you can see that the allowed deployment time can be changed by changing the deployment duration.
But in our [implementation](https://github.com/aurora-is-near/near-plugins/blob/add-upgrade-delay/near-plugins-derive/src/upgradable.rs#L124) the allowed timestamp to deploy the staged code can't be changed by changing the staging_duration. So to be able to deploy the code with the new duration you need first apply it by `up_apply_update_staging_duration` and then call again `up_stage_code`.

In other words, each of staging code and updating staging duration has its own flow with its own storage keys and don't affect each other.

birchmd: Ah, got it. Thanks for the correction @karim-en !

birchmd (approved):
I think we should get this merged. @mooori please take another look and let us know if you have any additional comments.

near-plugins-derive/src/upgradable.rs
karim-en: Renamed in c1654aea3df463524579d02fa8a091b8c59fb017

near-plugins-derive/src/upgradable.rs
karim-en: Reverted in c1654aea3df463524579d02fa8a091b8c59fb017

near-plugins-derive/src/upgradable.rs
mooori: When removing staged code, wouldn't it be more intuitive to remove `StagedTimestamp` instead of setting it to a new value?

mooori (commented):
LGTM in general. Left some minor comments that I think should be addressed before merging.

near-plugins/src/upgradable.rs
```diff
@@ -65,6 +67,23 @@ pub trait Upgradable {
...
+    /// Allows authorized account to apply the staging duration update.
+    fn up_apply_update_staging_duration(&self);
+}
+
+#[derive(Serialize)]
+pub struct UpgradableDurationStatus {
+    pub staging_duration: Option<near_sdk::Duration>,
+    pub staging_timestamp: Option<near_sdk::Timestamp>,
+    pub new_staging_duration: Option<near_sdk::Duration>,
+    pub new_staging_duration_timestamp: Option<near_sdk::Duration>,
```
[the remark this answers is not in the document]
karim-en: fixed f437e99ebe71dba00ded86caf42465d847aae435

near-plugins/src/upgradable.rs
```diff
@@ -65,6 +67,23 @@ pub trait Upgradable {
...
     fn up_deploy_code(&mut self) -> Promise;
+
+    /// Initialize the duration of the delay for deploying the staged code.
+    fn up_init_staging_duration(&self, staging_duration: near_sdk::Duration);
+
+    /// Allows authorized account to stage update of the staging duration.
+    fn up_stage_update_staging_duration(&self, staging_duration: near_sdk::Duration);
+
+    /// Allows authorized account to apply the staging duration update.
+    fn up_apply_update_staging_duration(&self);
```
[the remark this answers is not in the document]
karim-en: Fixed d1932d6de53cd099bdf00f4e5ec928a098228ede

near-plugins-derive/src/upgradable.rs
karim-en: Fixed 5c4fce0b5e25670a1d3810a00a08cfd566655486

mooori (approved):
@karim-en thanks!

mrLSD (approved):

Garrybrayn (commented):
l

Issue #48, opened 2022-12-07
Avoid duplication of test and example contracts


mooori:
## Description

For example for the `Pausable` plugin there is a basically identical `Counter` contract in the following locations:

- README.md
- near-plugins/src/pausable.rs (in `mod tests`)
- examples/pausable-examples/README.md
- examples/pausable-examples/pausable-base/src/lib.rs

## Problem

This duplication makes updates very cumbersome and it carries the risk of the contracts getting out of sync.

## Proposed solution

Have only _one_ contract per plugin which is used in tests and serves as example.

For instance the contract in `near-plugins/tests/contracts/access_controllable/` is compiled, deployed and interacted with by integration tests in `near-plugins/tests/access_controllable.rs`. This happens via `cargo test` and is hooked up with CI.

If this contract is polished I think we can remove the `AccessControllable` example in the README and instead add a link to this contract's `lib.rs`.

According to this pattern, contracts in `near-plugins/tests/contracts/` are integrated with `cargo test` and CI. Therefore I would argue no additional contracts are needed in `examples/**`. Instead examples link to `near-plugins/tests/contracts/` too.

In addition, unit tests in `/near-plugins/src/*.rs` can be replaced by the integration tests mentioned above.

birchmd:
I'm generally in favour of reducing code duplication. The rule of thumb I think about is whether the two pieces of code will change at the same time and for the same reasons. In the case of these tests and examples I think it is true they will always change together so it makes sense to unify them in my opinion.

Please proceed with your proposed solution @mooori

PR #51, opened 2022-12-20
test: add integration tests for `Pausable` plugins

mooori:
Previously `Pausable` was tested by unit tests via [`near_sdk::testing_env`](https://docs.rs/near-sdk/4.1.1/near_sdk/macro.testing_env.html), which mocks the chain.

The new integration tests compile the contract in `near-plugins/tests/contracts/pausable`, deploy it on chain via near `workspaces` and send transactions to it.

This is a step towards having _only one_ contract for the `Pausable` plugin which is used for tests and serves as documentation resp. example, as laid out in #48. In a follow up PR I’ll clean up duplicate example contracts.

## Other changes: refactor `near-plugins/tests/common`
- Removed superfluous types that complicated the code.
- Generalized helper functions to be used in tests of multiple plugins. Prior only `AccessControllable` was tested via integration tests.


near-plugins/tests/contracts/pausable/Makefile
```diff
@@ -0,0 +1,6 @@
+build:
```
birchmd, line 1: I think `build` and `expand` should be added as ["phony" targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) to following Makefile conventions.

near-plugins/tests/contracts/pausable/rust-toolchain
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "1.64.0"
```
birchmd, line 2: Could probably use 1.65.0 if we wanted to.

near-plugins/tests/pausable.rs
```diff
@@ -0,0 +1,562 @@
...
+    assert_insufficient_acl_permissions(
+        result,
+        "pa_pause_feature",
+        vec!["PauseManager".to_string()],
+    );
+}
+
+#[tokio::test]
+/// Only accounts that were granted a manager role may pause features.
+/// TODO same for unpause
```
birchmd, line 177: Remaining TODO item

birchmd (commented):
Looks pretty good to me, nice job! Just a few nits.

near-plugins/tests/pausable.rs
mooori: Sorry - done in 1b69596.

near-plugins/tests/contracts/pausable/Makefile
mooori: Done in 93887f3.

near-plugins/tests/contracts/pausable/rust-toolchain
mooori: Updating to `1.65.0` works fine locally on my machine, however it makes [CI fail](https://github.com/aurora-is-near/near-plugins/actions/runs/3747724193/jobs/6364726907). Perhaps CI cannot handle differing channels in [root](https://github.com/aurora-is-near/near-plugins/blob/b679c29fb45698d53cfed8abcc37a64664c57907/rust-toolchain) and test contract toolchains.

I openend #52 and would leave the channel as is for this PR.

near-plugins/tests/contracts/pausable/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+                "Method paused for large values of counter"
+            );
+        }
+
+        self.counter += 1;
+    }
+}
+
+/// Exposing internal methods to facilitate integration testing.
+/// TODO needed for this contract?
```
birchmd, line 118: Is this question answered now?

birchmd (approved):
LTGM, nice work

near-plugins/tests/contracts/pausable/src/lib.rs
mooori: It turned out that these functions are not needed for the `Pausable` test contract. They're removed in 5fe4f37.

Thanks for catching another open TODO!

mooori:
CI is green and I've double checked that there are no more open TODOs related to this PR, so merging now.

mrLSD (approved):

PR #57, opened 2023-01-05
refactor: names of methods and roles in the Acl contract

mooori:
The combination of role names and methods names in the test/demo contract for `AccessControllable` was confusing.

For the review I'd recommend looking at the contract (`lib.rs`) first, the changes in `tests/access_controllable.rs` should then be straightforward.


birchmd (approved):
LGTM thanks for this!

mrLSD (approved):

PR #66, opened 2023-01-16
test: add integration tests for `FullAccessKeyFallback`

mooori:
Adds integration tests and removes the corresponding unit tests. The background and motivation are described in https://github.com/aurora-is-near/near-plugins/pull/51, which does the same for `Pausable`.

This only ports existing unit tests. We'll probably want to add more tests. Before doing that, I propose to wait for `FullAccessKeyFallback` changes resulting from audits.


near-plugins/src/full_access_key_fallback.rs
```diff
@@ -42,47 +42,3 @@ impl AsEvent<FullAccessKeyAdded> for FullAccessKeyAdded {
         }
     }
 }
-
-#[cfg(not(target_arch = "wasm32"))]
-#[cfg(test)]
-mod tests {
-    // TODO: Make simulation test that verifies key get's added to the account
```
mooori, line 49: Resolved - this is asserted in the new integration tests.

near-plugins/tests/full_access_key_fallback.rs
```diff
@@ -0,0 +1,171 @@
...
+/// It returns a `near_sdk::PublicKey` since that's the type required for
+/// `FullAccessKeyFallback::attach_full_access_key`.
+fn new_public_key() -> near_sdk::PublicKey {
+    "ed25519:6E8sCci9badyRkXb3JoRpBj5p8C6Tw41ELDZoiihKEtp"
+        .parse()
+        .unwrap()
+}
+
+/// Converts a `near_sdk::PublicKey` to a `workspaces::types::PublicKey`.
+fn pk_sdk_to_workspaces(public_key: near_sdk::PublicKey) -> PublicKey {
```
birchmd, line 29: The implementation of this function is pretty hilarious. Serializing to JSON and back as a type conversion seems quite inefficient. Is there not another way to convert between these types? If not it's maybe worth raising as an issue in the workspaces repo.

near-plugins/tests/full_access_key_fallback.rs
```diff
@@ -0,0 +1,171 @@
...
+            .expect("Should view access keys");
+        assert_eq!(
+            access_keys.len(),
+            keys.len() + 1, // + 1 for the contract's key
+        );
+
+        // Assert the `access_keys` are the contract's key followed by `keys` (all full access).
+        let contract_key = self.contract.as_account().secret_key().public_key();
+        let expected_keys = iter::once(&contract_key).chain(keys.iter());
+        for (i, expected_key) in expected_keys.into_iter().enumerate() {
```
birchmd, line 104: Seems a little fragile to me that we rely on the specific order of the keys in this test. Is there a specific order that is expected from `view_access_keys`? If not then I think it would be better to compare these as sets instead.

near-plugins/tests/full_access_key_fallback.rs
mooori: `near-sdk` has `PublicKey` but no `SecretKey` type, so I think the conversion has to be via `PublicKey`. Looking at the [workspaces::types::PublicKey docs](https://docs.rs/workspaces/0.7.0/workspaces/types/struct.PublicKey.html#impl-Into%3CU%3E-for-PublicKey), I only see `Deserialize<'de>` for generating a `workspaces::types::PublicKey` from data.

Since this function is only available in tests, I wasn’t too concerned about the efficiency of the conversion. Though it’s definitely awkward. I’ve opened [this issue](https://github.com/near/workspaces-rs/issues/262) and left a comment in the code.

---

Perhaps there’s a more efficient way to do this conversion via `serde`, though not sure if it’s worth digging into that.

near-plugins/tests/full_access_key_fallback.rs
mooori: As [view_access_keys](https://docs.rs/workspaces/latest/workspaces/struct.Contract.html#method.view_access_keys) returns a list I was _assuming_ the returned keys are ordered. Since that’s not specified in the docs, you’re right, it’s not a solid approach.

It’s refactored to not rely on the order in 118de52. Currently it’s not possible to use a (`std`) set since `workspaces::types::PublicKey` doesn’t implement neither `Hash` nor `Ord`. I’ve opened [this issue](https://github.com/near/workspaces-rs/issues/263).

mooori:
It might take some time for fixes of above issues to land in `workspaces`, if they are fixed at all. While the issues lead to inefficient and awkward code, I think it’s not a blocker since this code is only available in tests. In that case we could move on with this PR.

WDYT @birchmd?

birchmd (approved):
Yes, I'm ok with merging this PR. Making issues on the workspaces repo is good enough to make sure we are being good open source citizens.

mrLSD (approved):

PR #69, opened 2023-01-23
chore: make Acl role limit explicit

mooori:


mooori:
CI failure is due to the PR being based on a external repo. Closing this in favor of #70, which is identical except that it’s based on a branch of this repo.

PR #70, opened 2023-01-23
chore: make Acl role limit explicit

mooori:
Currently, if the enum that represents Acl roles has too many variants, compilation fails with an [overflow error](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1fbdba773e8a7539e21cebdec62eb8d7) due to this type definition:

https://github.com/aurora-is-near/near-plugins/blob/dbec4424c5bbfad5fedbfcbe877ed16bacfa18f3/near-plugins-derive/src/access_control_role.rs#L194-L198

This might be hard to understand for developers using the plugin. The changes in this PR produce a more helpful error in case there are too many enum variants. In addition, an upcoming PR will rely on that limit, so better to have it explicitly defined.

### DIFF
--- a/near-plugins-derive/src/access_control_role.rs
+++ b/near-plugins-derive/src/access_control_role.rs
@@ -46,6 +46,13 @@ use quote::quote;
 use std::convert::TryFrom;
 use syn::{parse_macro_input, ItemEnum};
 
+/// Roles as are represented by enum variants which are, in turn, represented by
+/// `u128` bitflags. Each variant requires two flags, one for the role itself
+/// and one for the corresponding admin permission. This would allow for 64
+/// roles. However, one flag is reserved for `__SUPER_ADMIN`, leaving 127
+/// bits that can fit 63 roles.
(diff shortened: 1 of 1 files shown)


near-plugins-derive/src/access_control_role.rs
```diff
@@ -46,6 +46,13 @@ use quote::quote;
 use std::convert::TryFrom;
 use syn::{parse_macro_input, ItemEnum};
 
+/// Roles as are represented by enum variants which are, in turn, represented by
+/// `u128` bitflags. Each variant requires two flags, one for the role itself
+/// and one for the corresponding admin permission. This would allow for 64
+/// roles. However, one flag is reserved for `__SUPER_ADMIN`, leaving 127
+/// bits that can fit 63 roles.
+pub const MAX_ROLE_VARIANTS: u8 = 63;
```
mooori, line 54: Making it `pub` to have it surface in docs, since the assert message below mentions it.

birchmd (approved):

mrLSD (approved):

PR #71, opened 2023-01-24
docs: add notes on upgrading vulnerable code

mooori:
- Describes how contracts with vulnerable code might mitigate risks when using `Upgradable`.
- Does some language reworks of preceding sections (doing that in a separate PR might be too much overhead).

### DIFF
--- a/near-plugins/src/upgradable.rs
+++ b/near-plugins/src/upgradable.rs
@@ -22,6 +22,27 @@
 //!
 //! After the code is deployed, it should be removed from staging. This will prevent old code with a
 //! security vulnerability to be deployed.
+//!
+//! ## Upgrading code that contains a security vulnerability
+//!
+//! Once code is staged for an upgrade, it is publicly visible via [`Upgradable::up_staged_code`].
+//! Staged code that fixes a security vulnerability might be discovered by an attacker who then
+//! exploits the vulnerability before its fix is deployed.
+//!
+//! To avoid that, the upgrade can be executed by calling [`Upgradable::up_stage_code`] and
+//! [`Upgradable::up_deploy_code`] in a [batch transaction]. Since [`Upgradable::up_deploy_code`]
+//! returns a promise that ultimately deploys the new contract code, a theoretical risk remains.
+//! However, the [time between scheduling and execution] of a promise hardly allows an attacker to
+//! exploit a vulnerability: they would have to retrieve the bytes of the staged code, reverse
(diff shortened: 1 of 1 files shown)


near-plugins/src/upgradable.rs
```diff
@@ -1,24 +1,48 @@
...
+//! example if an upgrade requires migration or initialization. In that case, it is recommended to
+//! run a batched transaction where [`Upgradable::up_deploy_code`] is called first, and then a
 //! function that executes the migration or initialization.
 //!
-//! After the code is deployed, it should be removed from staging. This will prevent an old code
-//! with a security vulnerability to be deployed, in case it was upgraded using other mechanism.
+//! ## Stale staged code
+//!
+//! After the code is deployed, it should be removed from staging. This will prevent old code with a
+//! security vulnerability to be deployed.
```
birchmd, line 24: Is there a method to remove the staged code? If this is something that should happen every time should we make the upgrade implementation remove it automatically?

birchmd (approved):

near-plugins/src/upgradable.rs
mooori: I’m merging now with that paragraph as is, since it’ll be rewritten soon in upcoming PRs (see below).

---

Staged code can be removed by passing an empty vector to `up_stage_code` ([ref](https://github.com/aurora-is-near/near-plugins/blob/c043add4c2a0810872c4326a55c5162bccd3f4ee/near-plugins-derive/src/upgradable.rs#L32-L39)). It’s currently not mentioned in docs and unidiomatic, I think. For that behavior it is expected to be wrapped in an `Option`? In that case I can open an issue to change the signature of `up_stage_code` to take `code: Option<Vec<u8>>`. Also it would be consistent with `up_staged_code` [returning None](https://github.com/aurora-is-near/near-plugins/blob/c043add4c2a0810872c4326a55c5162bccd3f4ee/near-plugins/src/upgradable.rs#L36-L37) if there’s no code staged.

Staged code should be removed automatically after deployment, it was also pointed out [here](https://github.com/AuditoneCodebase/Aurora-audit-review/issues/9). I’ll make a PR for that change soon.

birchmd: Passing in an `Option` to allow manual unstaging sounds good to me. And it's good we'll be adding the automatic unstaging after deploy as well.

mooori: Now there's #73 to track this.

mrLSD (approved):

PR #74, opened 2023-01-25
docs: describe migration strategy for Upgradable

mooori:
Since an upgrade that requires migration of contract state can be tricky.

### DIFF
--- a/near-plugins/src/upgradable.rs
+++ b/near-plugins/src/upgradable.rs
@@ -13,10 +13,21 @@
 //! ## Permissions
 //!
 //! Only an authorized account is allowed to call [`Upgradable::up_stage_code`] and
-//! [`Upgradable::up_deploy_code`]. There may be several reasons to protect `deploy_code`. For
-//! example if an upgrade requires migration or initialization. In that case, it is recommended to
-//! run a batched transaction where [`Upgradable::up_deploy_code`] is called first, and then a
-//! function that executes the migration or initialization.
+//! [`Upgradable::up_deploy_code`]. There may be several reasons to protect `deploy_code`, one of
+//! them being an upgrade that requires migration.
+//!
+//! ## Upgrade that requires state migration
+//!
+//! If a contract upgrade requires state migration, it is recommended to execute the upgrade in a
+//! batch transaction that calls [`Upgradable::up_deploy_code`] followed by the migration method. To
+//! verify the migration was successful, in addition, a function that deserializes contract state
+//! might be called.
+//!
+//! Note that even if above actions are executed in a batch transaction, the staged code is still
+//! deployed in case the migration fails. This is due to the [asynchronous nature] of NEAR and the
+//! deployment ultimately being executed in a separate promise that is not affected by the outcome
+//! of the batch transaction. Still, a batch transaction helps to minimize the time between the
+//! deployment of new code and the migration of state.
 //!
 //! ## Stale staged code
 //!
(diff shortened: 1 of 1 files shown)


near-plugins/src/upgradable.rs
```diff
@@ -13,10 +13,21 @@
 //! ## Permissions
 //!
 //! Only an authorized account is allowed to call [`Upgradable::up_stage_code`] and
-//! [`Upgradable::up_deploy_code`]. There may be several reasons to protect `deploy_code`. For
-//! example if an upgrade requires migration or initialization. In that case, it is recommended to
```
mooori, line 17: Reasoning for removing "initialization" from this paragraph: All `Upgradable` methods have parameter `self`, so they require the contract to be already initialized.

near-plugins/src/upgradable.rs
```diff
@@ -13,10 +13,21 @@
...
+//!
+//! ## Upgrade that requires state migration
+//!
+//! If a contract upgrade requires state migration, it is recommended to execute the upgrade in a
+//! batch transaction that calls [`Upgradable::up_deploy_code`] followed by the migration method. To
+//! verify the migration was successful, in addition, a function that deserializes contract state
+//! might be called.
+//!
+//! Note that even if above actions are executed in a batch transaction, the staged code is still
+//! deployed in case the migration fails. This is due to the [asynchronous nature] of NEAR and the
```
birchmd, line 27: I'm not sure that this is right. When an account calls itself a "local receipt" is created and such receipts can be executed in the same block, so it's possible that the whole batch is atomic. Maybe we should write a test to verify this one way or the other?

mooori: Interesting point and curiously [this doc](https://docs.near.org/concepts/basics/transactions/overview) doesn’t mention local receipts. Therefore I assumed [this paragraph](https://docs.near.org/concepts/basics/transactions/overview#atomicity) is applicable for `up_deploy_code` since it spawns a [new promise](https://github.com/aurora-is-near/near-plugins/blob/1cf0aa61e38d3b02803642983795c927598375ba/near-plugins-derive/src/upgradable.rs#L53-L54). If I understand correctly, this means the receipt to deploy the staged code is not created during the initial transaction to receipt conversion.

Then I stumbled on [this SO answer](https://stackoverflow.com/a/67708951/) (I suppose from a nearcore developer) which states:

> The local receipt can only be created from a conversion of a transaction to a receipt.

Without digging into the implementation, it seems difficult to find docs on local receipts. Which makes me wonder if we had a test, would it be safe to assume the observed behavior _always_ occurs? If not, perhaps it would be safer to rephrase the highlighted paragraph and mention these concerns?

birchmd: Ah, I see, thanks for the SO reference.

It seems sub-optimal to me that a failed migration could leave a contract is a totally broken state and all we do as mitigation is warn people this might happen. I wonder if it makes sense to extend the upgrade plugin to include an optional string which is the name of the migration method. Then the promise produced by `up_deploy_code` could itself be a batch which includes both the deploy and the migration call. This would ensure that the migration and deploy are always atomic.

mooori: >  I wonder if it makes sense to extend the upgrade plugin to include an optional string which is the name of the migration method.

Sounds like a great way to avoid that problem and improve the `Upgradable` plugin. I've opened #76 to track it.

That change will come with a rewrite of the docs changed here, so for now I think it makes sense to leave them as is and close this PR.

PR #78, opened 2023-02-02
feat(AccessControllable): add `acl_get_permissioned_accounts`

mooori:
Adds method `acl_get_permissioned_accounts` which returns information about _all_ outstanding permissions.

This method will be available on every contract that implements `AccessControllable`.

### DIFF
--- a/near-plugins/tests/access_controllable.rs
+++ b/near-plugins/tests/access_controllable.rs
@@ -4,10 +4,12 @@ pub mod common;
 
 use common::access_controllable_contract::AccessControllableContract;
 use common::utils::{
-    assert_insufficient_acl_permissions, assert_private_method_failure, assert_success_with,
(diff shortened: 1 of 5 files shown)


near-plugins/tests/access_controllable.rs
```diff
@@ -117,6 +119,82 @@ async fn call_increase_2(
         .await
 }
 
+/// Returns new `PermissionedAccounts` for [`ALL_ROLES`].
+fn new_permissioned_accounts() -> PermissionedAccounts {
```
birchmd, line 123: Does it make sense to derive `Default` on `PermissionedAccounts` instead?

birchmd (approved):

near-plugins/tests/access_controllable.rs
birchmd: Ah, nevermind `ALL_ROLES` is a test-specific constant.

mrLSD (approved):

PR #79, opened 2023-02-09
test: add integration tests for `Upgradable`

mooori:


near-plugins/tests/upgradable.rs
```diff
@@ -0,0 +1,681 @@
...
+
+    // Deploy staged code.
+    let res = setup.upgradable_contract.up_deploy_code(&owner).await?;
+    assert_success_with_unit_return(res);
+
+    Ok(())
+}
+
+/// TODO stage code that corresponds to a valid contract. Call a method of that contract to verify
+/// the staged code was deployed.
```
mooori, line 383: Carrying over [this TODO](https://github.com/aurora-is-near/near-plugins/blob/0c25fab016096792112798d1ace1e2e5ff598b3a/near-plugins/src/upgradable.rs#L132). I suggest to tackle it in a separate PR.

near-plugins/tests/common/utils.rs
```diff
@@ -146,3 +160,44 @@ pub fn assert_failure_with(res: ExecutionFinalResult, must_contain: &str) {
...
+/// forwarding provided by this function is reasonly fast in our tests for durations that correspond
+/// to less than 100 seconds.
+///
+/// [Time travels]: https://github.com/near/workspaces-rs#time-traveling
+pub async fn fast_forward_beyond(worker: &Worker<Sandbox>, duration: Duration) {
+    let initial_timestamp = block_timestamp(worker).await;
+
+    // Estimating a number of blocks to skip based on `duration` and calling `fast_forward` only
+    // once seems more efficient. However, that leads to jittery tests as `fast_forward` may _not_
+    // forward the block timestamp significantly.
```
birchmd, line 196: I feel like there's a middle ground here where we can fast-forward in larger steps until the duration to wait is smaller. For example, if we are skipping 10s into the future then it seems safe that we could skip 5 blocks and then check how much farther we need to go.

But then again, this is only used in tests so you're probably right that it is not worth the additional complexity for a minor speed-up. Let's leave it as-is unless it becomes a problem.

birchmd (approved):
LGTM, nice work

near-plugins/tests/common/utils.rs
mooori: It’s tricky because we need to move forward in time but `fast_forward` moves forward in blocks. Since forwarding by multiple blocks still may _not_ forward the timestamp, I guess finding that middle ground is difficult. That’s why I went for this very naive approach.

mrLSD (approved):

Garrybrayn:
image_1709488019463306

PR #82, opened 2023-02-15
ci: switch to self hosted runners

mooori:
Run CI jobs on self hosted runners instead of runners provided by Github.


mooori:
Closing: This repo is public, so there's no need to use self-hosted runners for actions:

> GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.

[Source](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions)

PR #83, opened 2023-02-15
test(Upgradable): transact with contract after deployment

mooori:
Adds a test sending a transaction to a contract that was upgraded using `Upgradable`. The new version of the contract is located in `near-plugins-derive/tests/contracts/upgradable_2`.


birchmd (approved):
Nice work!

mrLSD (approved):

PR #84, opened 2023-02-21
ACL: Remove `__acl` field

karim-en:
Currently, the ACL plugins can't be added to already deployed contracts without migration due to the `__acl` field.
This pull request removes the `__acl` field and uses the storage key instead.

Added internal/private API:
```
fn acl_get_storage(&self) -> Option<#acl_type>
fn acl_get_or_init(&mut self) -> #acl_type
fn acl_init_storage_unchecked(&mut self)
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 Cargo.lock
 near-plugins-derive/tests/contracts/*/target
 examples/target
-
+**/target/**
 # Ignore IDE data
 .vscode/
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 /target
 Cargo.lock
 near-plugins-derive/tests/contracts/*/target
+near-plugins/tests/contracts/*/target
 examples/target
-**/target/**
+
 # Ignore IDE data
 .vscode/
 .idea/
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 /target
 Cargo.lock
 near-plugins-derive/tests/contracts/*/target
-near-plugins/tests/contracts/*/target
 examples/target
 
 # Ignore IDE data
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -108,11 +108,11 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 })
             }
 
-            fn acl_get_or_init(&self) -> #acl_type {
+            fn acl_get_or_init(&mut self) -> #acl_type {
                 self.acl_get_storage().unwrap_or_else(|| self.acl_init_storage_unchecked())
             }
 
-            fn acl_init_storage_unchecked(&self) -> #acl_type {
+            fn acl_init_storage_unchecked(&mut self) -> #acl_type {
                 let base_prefix = <#ident as AccessControllable>::acl_storage_prefix();
                 let acl_storage: #acl_type = Default::default();
                 near_sdk::env::storage_write(
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -84,7 +84,7 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
             Permissions,
             Bearers,
             BearersSet { permission: #bitflags_type },
-            AclField,
+            AclStorage,
         }
 
         /// Generates a prefix by concatenating the input parameters.
@@ -100,10 +100,10 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 let base_prefix = <#ident as AccessControllable>::acl_storage_prefix();
                 near_sdk::env::storage_read(&__acl_storage_prefix(
                     base_prefix,
-                    __AclStorageKey::AclField,
+                    __AclStorageKey::AclStorage,
                 ))
-                .map(|acl_field_bytes| {
-                    #acl_type::try_from_slice(&acl_field_bytes)
+                .map(|acl_storage_bytes| {
+                    #acl_type::try_from_slice(&acl_storage_bytes)
                         .unwrap_or_else(|_| near_sdk::env::panic_str("ACL: invalid acl storage format"))
                 })
             }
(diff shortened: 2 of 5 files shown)


birchmd (approved):
Looks reasonable to me!

mooori (commented):
Thanks for this change! Initially the acl plugin used `near_sdk::collections`  types which couldn’t be used with `near_sdk::env::storage_{read, write}`. Great to see it is possible with `near_sdk::store` types.

---

Why is it necessary to explicitly initiate storage with `acl_init_storage()`? Couldn’t storage be init'ed implicitly if needed? For instance instead of `acl_get` there could be `acl_get_or_init`:

```rust
fn acl_get_or_init(&mut self) -> #acl_type {
    // Assume `acl_init_storage_unchecked` returns the struct it wrote to storage.
    self.acl_get_storage().unwrap_or_else(|| self.acl_init_storage_unchecked())
}
```

This is in line with Acl initiating data structures if needed ([example](https://github.com/aurora-is-near/near-plugins/blob/fec96ab199858cdc3d4f6b07750f9fd5e794c04f/near-plugins-derive/src/access_controllable.rs#L110-L112), [example](https://github.com/aurora-is-near/near-plugins/blob/fec96ab199858cdc3d4f6b07750f9fd5e794c04f/near-plugins-derive/src/access_controllable.rs#L391-L393)) and also other plugins don’t require `*_init_storage`. Moreover, I think it has the following benefits:

### No panics
In the current implementation `acl_get` may panic and it is called in the implementation of almost every `AccessControllable` trait method. This means we must explain to users when/why Acl may panic due to storage not being init’ed. With `acl_get_or_init` this wouldn’t be necessary since there is no panic.

### No need to explicitly init storage
Which makes the plugin easier to use.

### No need for `acl_init_storage` in the public API
Since it is init’ed internally if needed.

### Note
If going with above proposal, methods that don’t take `&mut self` cannot call `acl_get_or_init(&mut self)`. Instead they can do something like:

```rust
fn acl_has_role(&self, role: String, account_id: ::near_sdk::AccountId) -> bool {
    let role: #role_type = ::std::convert::TryFrom::try_from(role.as_str()).unwrap_or_else(|_| ::near_sdk::env::panic_str(#ERR_PARSE_ROLE));
    match self.acl_get_storage() {
        Some(acl) => acl.has_role(role, &account_id),
        None => false
    }
}
```

.gitignore
```diff
@@ -2,7 +2,7 @@
 Cargo.lock
 near-plugins-derive/tests/contracts/*/target
 examples/target
-
+**/target/**
```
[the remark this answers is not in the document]
karim-en: Changed to `near-plugins/tests/contracts/*/target` in 1c504a7e08e2f7e0c6382f9d4a521f3a13c5998c

near-plugins-derive/src/access_controllable.rs
```diff
@@ -90,6 +84,7 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
             Permissions,
             Bearers,
             BearersSet { permission: #bitflags_type },
+            AclStorage,
```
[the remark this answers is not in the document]
karim-en: According to the rules, it is dangerous to the storage key order. Also, we already have deployed contracts on mainnet that use `acl` plugin, so it will complicate the migration.

To migrate the existing contracts we will just declare the contract structure like that :
```
#[derive(BorshDeserialize, BorshSerialize)]
pub struct CounterV0 {
    counter: u64,
    __acl: __Acl,
}
```

The `Permissions` and `Bearers` maps will be migrated automatically without any extra steps.

karim-en:
> https://github.com/aurora-is-near/near-plugins/pull/84#pullrequestreview-1315085648

@mooori @birchmd 

Personally, I prefer panic instead of returning default values because normally it shouldn't panic if the contract was implemented correctly, so it indicates there is some bug in the initialization or migration code.

But anyway I've implemented this proposal 35635bc6e6e4c9ec7fcf4383ab9d183034809ff0, to be able to merge this pull request asap.

.gitignore
```diff
@@ -1,6 +1,7 @@
 /target
 Cargo.lock
 near-plugins-derive/tests/contracts/*/target
+near-plugins/tests/contracts/*/target
```
mooori, line 4: This directory doesn't exist anymore in the repo, `near-plugins/tests` has been moved to `near-plugins-derive/tests`. Probably there are some stale files on your machine (fix `rm -r near-plugins/tests`).

near-plugins-derive/src/access_controllable.rs
```diff
@@ -100,6 +95,34 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
...
+                    #acl_type::try_from_slice(&acl_storage_bytes)
+                        .unwrap_or_else(|_| near_sdk::env::panic_str("ACL: invalid acl storage format"))
+                })
+            }
+
+            fn acl_get_or_init(&self) -> #acl_type {
+                self.acl_get_storage().unwrap_or_else(|| self.acl_init_storage_unchecked())
+            }
+
+            fn acl_init_storage_unchecked(&self) -> #acl_type {
```
mooori, line 115:
```suggestion
            fn acl_init_storage_unchecked(&mut self) -> #acl_type {
```

It modifies state.

near-plugins-derive/src/access_controllable.rs
```diff
@@ -100,6 +95,34 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
...
+                    base_prefix,
+                    __AclStorageKey::AclStorage,
+                ))
+                .map(|acl_storage_bytes| {
+                    #acl_type::try_from_slice(&acl_storage_bytes)
+                        .unwrap_or_else(|_| near_sdk::env::panic_str("ACL: invalid acl storage format"))
+                })
+            }
+
+            fn acl_get_or_init(&self) -> #acl_type {
```
mooori, line 111:
```suggestion
            fn acl_get_or_init(&mut self) -> #acl_type {
```

It may modify state.

mooori (commented):
LGTM, just some functions should mutably borrow `self`, I think (see below).

This sentence is outdated:

https://github.com/aurora-is-near/near-plugins/blob/35635bc6e6e4c9ec7fcf4383ab9d183034809ff0/near-plugins-derive/tests/contracts/access_controllable/src/lib.rs#L124-L125

Possible replacement:

```rust
/// Most of them are implemented for the type that holds the plugin's state,
/// which can be accessed with `self.acl_get_or_init()`.
```

birchmd (approved):
RE: panic or not to panic

Probably the best solution is to have both functions and let developers decide what makes sense for them. This would be analogous to how Near SDK allows your contract to derive `PanicOnDefault` or not.

Making panicking on missing storage configurable would be a more complex change so I don't think we should do it as part of this PR, but maybe it is something we want to implement in the future.

.gitignore
karim-en: 59245225e53f1deee0c14a5c67e69b9cb5d5346a

mooori (approved):
Thanks!

PR #85, opened 2023-03-02
feat(Upgradable)!: use Acl for authorization 

mooori:
Previously `Upgradable` authorization was handled by `Ownable`. An explanation of how to use `Upgradable` with Acl is available in the [example contract](https://github.com/aurora-is-near/near-plugins/blob/d5096361fa42ea21b82f0d071497400f357919eb/near-plugins-derive/tests/contracts/upgradable/src/lib.rs).

# Acl Roles
Use cases which don’t require fine grained permissions can define only a single role and whitelist it for all `Upgradable` features. See `Role::DAO` in the example contract.

The other roles show how to achieve more fine grained permissions by whitelisting a role only for a subset of the `Upgradable` features.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,13 +61,13 @@ Documentation of all methods provided by `Pausable` is available in the [definit
 
 ### [Upgradable](/near-plugins/src/upgradable.rs)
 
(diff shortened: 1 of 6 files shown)


birchmd (approved):
Looks good to me. Nice job!

mrLSD (approved):

PR #86, opened 2023-03-07
feat(Upgradable)!: enable batched fn call after deploy

### LINKED ISSUE #76: `Upgradable`: don't deploy code if state migration fails
> It seems sub-optimal to me that a failed migration could leave a contract is a totally broken state and all we do as mitigation is warn people this might happen. I wonder if it makes sense to extend the upgrade plugin to include an optional string which is the name of the migration method. Then the promise produced by `up_deploy_code` could itself be a batch which includes both the deploy and the migration call. This would ensure that the migration and deploy are always atomic.

> Edit: creating a batch promise with both deploy and migration is [how we do it in Aurora Engine](https://github.com/aurora-is-near/aurora-engine/blob/master/engine-sdk/src/near_runtime.rs#L53) for example.

_Originally posted by @birchmd in https://github.com/aurora-is-near/near-plugins/pull/74#discussion_r1086796106_

mooori:
Closes #76 and this [external issue](https://github.com/AuditoneCodebase/Aurora-audit-review/issues/27).

Adds an additional parameter to `Upgradable::up_deploy_code` that allows to optionally specify a function call to be batched with the code deployment. This might be a function that migrates state. If that function call fails, the deployment is rolled back and the old code remains active.

# Breaking change
This is a breaking change since it modifies the signature of `Upgradable::up_deploy_code`. Users that want to maintain the previous behavior (deployment without function call) can migrate in the following way:

```rust
// Previously:
up_deploy_code()

// Same behavior with the new API:
up_deploy_code(None)
```

### DIFF
--- a/near-plugins-derive/src/upgradable.rs
+++ b/near-plugins-derive/src/upgradable.rs
@@ -180,7 +180,7 @@ pub fn derive_upgradable(input: TokenStream) -> TokenStream {
             }
 
             #[#cratename::access_control_any(roles(#(#acl_roles_code_deployers),*))]
-            fn up_deploy_code(&mut self) -> near_sdk::Promise {
+            fn up_deploy_code(&mut self, function_call_args: Option<#cratename::upgradable::FunctionCallArgs>) -> near_sdk::Promise {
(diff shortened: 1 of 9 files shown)


birchmd (approved):
LGTM!

mrLSD (approved):

PR #88, opened 2023-03-08
docs(Upgradable): code removal after deplyoment

mooori:
Makes it more explicit how staged code can be removed after it was deployed. Mentions the benefit of freeing storage.

Related to [this issue](https://github.com/AuditoneCodebase/Aurora-audit-review/issues/9) and its comment.


birchmd (approved):
The change itself looks good to me. Some other random thoughts about this:

- How much gas does it actually cost to remove an entry from storage? If it is not much then maybe this argument about saving gas is not as relevant.
- Should we recommend if users want to atomically update and remove then they could submit a batch transaction that does both?
- Should we recommend that developers ensure upgrades are always idempotent since it is theoretically possible an upgrade could be applied multiple times if it is not removed immediately?

birchmd (approved):
Looks good to me. Thanks for the clarifications @mooori !

mrLSD (approved):

PR #92, opened 2023-03-13
refactor!: avoid cloning storage prefixes

mooori:


near-plugins/src/ownable.rs
```diff
@@ -25,7 +25,7 @@ pub trait Ownable {
     /// #[ownable(owner_storage_key="CUSTOM_KEY")]
     /// struct Contract { /* ... */}
     /// ```
-    fn owner_storage_key(&self) -> Vec<u8>;
+    fn owner_storage_key(&self) -> &'static [u8];
```
birchmd, line 28: I wonder if we should use `&[u8]` instead of enforcing a static lifetime. I can't really think of any, but maybe there is some use case where the storage key is not known at compile-time for some reason?

mooori: @birchmd Can we merge this as is in order to fix the external issue? I would suggest addressing the change to return `&[u8]` in a separate issue/PR due to the issue described below.

---

Returning `&[u8]` is the better option and it works for all traits except `AccessControllable`. For the other traits, the function which returns the storage prefix has parameter `&self`, so returning `&[u8]` (without lifetime specifier) is fine.

As of now, `AccessControllable::acl_storage_prefix` cannot take `&self` due to [this](https://github.com/aurora-is-near/near-plugins/blob/8853caf0015b922280f0846644e2684ecd9d4944/near-plugins-derive/src/access_controllable.rs#L117) usage of `Default::default()`. Therefore the return value of `acl_storage_prefix` must have a lifetime specifier. I think we have the following options:

- Refactor the default implementation of `AccessControllable` to not rely on `Default`. Then return `&[u8]` from all trait methods that return the storage prefix.
- Migrate the other traits to return `&[u8]` and accept `AccessControllable` as an outlier whose storage prefix function returns `&'static [u8]`.
- Defer this change until a use case pops up where returning `&'static [u8]` is not possible.

What do you think?

birchmd (approved):

near-plugins/src/ownable.rs
mooori: Tracked by issue #93.

mrLSD (approved):

PR #94, opened 2023-03-14
ACL: Add transfer super admin api

karim-en:
Currently, the ACL plugin provides api to init single super admin by `acl_init_super_admin`. But to transfer the admin to another account, the developers need to write their own method, like this:
```
    pub fn transfer_super_admin(&mut self, account_id: AccountId) {
        let current_super_admin = env::predecessor_account_id();

        near_sdk::require!(
            self.acl_get_or_init()
                .revoke_super_admin_unchecked(&current_super_admin),
            "Failed to revoke super-admin."
        );

        near_sdk::require!(
            self.acl_get_or_init().init_super_admin(&account_id),
            "Failed to init super-admin."
        );
    }
```

This pull request adds a default implementation to transfer the super admin permissions to another account, this could be useful for transferring the super admin to DAO or another account, or on migration from one DAO to another.

Added pubic APIs:
```
fn acl_transfer_super_admin(&mut self, account_id: AccountId) -> Option<bool>;
```

### DIFF
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -194,6 +194,16 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 Some(self.revoke_super_admin_unchecked(account_id))
             }
 
+            fn transfer_super_admin(&mut self, account_id: &::near_sdk::AccountId) -> Option<bool> {
+                let current_super_admin = ::near_sdk::env::predecessor_account_id();
+                if !self.is_super_admin(&current_super_admin) {
+                    return None;
+                }
+
+                self.revoke_super_admin_unchecked(&current_super_admin);
+                Some(self.add_super_admin_unchecked(&account_id))
+            }
+
             /// Revokes super-admin permissions from `account_id` without checking any
             /// permissions. It returns whether `account_id` was a super-admin.
             fn revoke_super_admin_unchecked(&mut self, account_id: &::near_sdk::AccountId) -> bool {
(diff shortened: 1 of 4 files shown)


near-plugins-derive/tests/access_controllable.rs
```diff
@@ -397,6 +397,57 @@ async fn test_acl_revoke_super_admin() -> anyhow::Result<()> {
...
+    // Transfer is a no-op if caller is not a super-admin.
+    let res = setup
+        .contract
+        .acl_transfer_super_admin(&caller_unauth, super_admin.id())
+        .await?;
+    assert_eq!(res, None);
+    setup
+        .contract
+        .assert_acl_is_super_admin(true, setup.contract_account(), super_admin.id())
+        .await;
```
mooori, line 423:
```suggestion
    setup
        .contract
        .assert_acl_is_super_admin(true, setup.contract_account(), super_admin.id())
        .await;
    setup
        .contract
        .assert_acl_is_super_admin(false, setup.contract_account(), new_super_admin.id())
        .await;
```

mooori (commented):
Thanks for this devx improvement! Docs should be changed slightly and I would add another assertion in the test. Otherwise LGTM.

birchmd (approved):
Thanks @karim-en !

mooori (approved):

near-plugins-derive/src/access_controllable.rs
```diff
@@ -194,6 +194,16 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
...
 
+            fn transfer_super_admin(&mut self, account_id: &::near_sdk::AccountId) -> Option<bool> {
+                let current_super_admin = ::near_sdk::env::predecessor_account_id();
+                if !self.is_super_admin(&current_super_admin) {
+                    return None;
+                }
+
+                self.revoke_super_admin_unchecked(&current_super_admin);
+                Some(self.add_super_admin_unchecked(&account_id))
+            }
```
sept-en, line 205: I suggest that we first add the super admin, and only if this is successful, then revoke the super admin. Generally, it's a better approach, especially if something will eventually be changed in these two functions on how they work. This is required to make sure the contract doesn't accidentally become locked on ACL.

CC @karim-en @mooori

mooori: The state modifications in `acl_transfer_super_admin` should be atomic since no promises or cross contract calls are scheduled. So I think currently it’s not possible to lock Acl in that way. I assume it’s unlikely that the default `AccessControllable` implementation will schedule promises or xcc in the future. Still I see the point of making this as future proof as possible. I can prepare a PR to reorder it tomorrow.

mooori: #101 reorders the operations.

mrLSD (approved):

mrLSD (approved):

PR #101, opened 2023-03-16
refactor(Acl): reorder state modifications

mooori:
Reorders the state modifications in `acl_transfer_super_admin` as suggested in [this comment](https://github.com/aurora-is-near/near-plugins/pull/94#discussion_r1137435879).

### DIFF
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -207,8 +207,17 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                     return None;
                 }
 
+                // The following state modifications can be considered atomic: They happen in the
+                // current `FunctionCall` action because no promises or cross contract calls are
+                // scheduled.
+                //
+                // If this ever changes, it should be avoided that revoking `current_super_admin`
+                // succeeds and adding `account_id` as super-admin fails. This could lock contracts
+                // in a state without super-admins. To protect against this scenario, the new
+                // super-admin is added first.
+                let is_new_super_admin = self.add_super_admin_unchecked(&account_id);
                 self.revoke_super_admin_unchecked(&current_super_admin);
-                Some(self.add_super_admin_unchecked(&account_id))
+                Some(is_new_super_admin)
             }
 
             /// Revokes super-admin permissions from `account_id` without checking any
--- a/near-plugins-derive/src/access_controllable.rs
+++ b/near-plugins-derive/src/access_controllable.rs
@@ -215,7 +215,14 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
                 // succeeds and adding `account_id` as super-admin fails. This could lock contracts
                 // in a state without super-admins. To protect against this scenario, the new
                 // super-admin is added first.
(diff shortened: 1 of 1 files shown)


near-plugins-derive/src/access_controllable.rs
```diff
@@ -207,8 +207,17 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
...
+                // The following state modifications can be considered atomic: They happen in the
+                // current `FunctionCall` action because no promises or cross contract calls are
+                // scheduled.
+                //
+                // If this ever changes, it should be avoided that revoking `current_super_admin`
+                // succeeds and adding `account_id` as super-admin fails. This could lock contracts
+                // in a state without super-admins. To protect against this scenario, the new
+                // super-admin is added first.
+                let is_new_super_admin = self.add_super_admin_unchecked(&account_id);
                 self.revoke_super_admin_unchecked(&current_super_admin);
```
birchmd, line 219: I think this needs to be under an `if` statement now (i.e. only revoke the current super admin if the new one is successfully added). Otherwise there could be a problem where there are zero super admins at the end of the operation (the new one failed to add but we removed the current one anyway). In particular, I think this bug would manifest if you tried to transfer the super admin to the same account that is already the super admin.

mooori: > the new one failed to add

The return value indicates if it `account_id` is a _new_ super-admin:

https://github.com/aurora-is-near/near-plugins/blob/4f1c8d19f29113c5e27ad841803f8e2722ceb543/near-plugins-derive/src/access_controllable.rs#L161-L165

So if it doesn't panic, then `account_id` is super-admin after it returns. Assuming transferring to an `account_id` that is already super-admin is valid (although a bit awkward), I think the `if` isn’t needed.

---

I just noticed that this implementation removes `account_id` if it equals `predecessor_account_id()`. So if Bob is the only super-admin and passes his own account id to `acl_transfer_super_admin`, then there would be no super-admin left. That is even more awkward, but still probably we should protect against it with something like `assert!(current_super_admin != account_id)`?

birchmd: > So if Bob is the only super-admin and passes his own account id to acl_transfer_super_admin, then there would be no super-admin left

Yes, that's what I said in my comment. And as I said in that comment I think the way to avoid this bug is to have

```rust
if is_new_super_admin {
    self.revoke_super_admin_unchecked(&current_super_admin);
}
```

birchmd: Ah, but I see what you are saying. You mean that if Alice and Bob are both super admins and Bob calls `acl_transfer_super_admin(Alice)` then in my version nothing happens (Bob is not removed as a super admin), which is maybe counter-intuitive.

So I agree that what we need here is instead an early return in the case that `predecessor_id == account_id` (i.e. self-transfer is a no-op). Panic could also be acceptable instead of early return, but I think just returning `Some(true)` is also fine.

mooori: > I think just returning Some(true) is also fine

That seems intuitive to me and I think for a plugin like this it is preferable to avoid panics. The early return is added in b2df328.

sept-en (approved):

mrLSD (approved):

PR #102, opened 2023-03-27
Fulfill TODO

SteMak:
The ToDo refers to PR #12.
As it was merged, the code can be updated there.

In such a way, some unnecessary actions are performed (`.iter().skip(0).take(1).cloned().collect()`).
However, the code looks clearer.

The change does not break any tests.

If the PR would not be accepted, the ToDo should be reworked or removed.


mooori:
> The CI failure seems to be unrelated to the changes made in this PR. We are investigating it.

CI is passing now after an issue with its setup has been resolved.

mrLSD (approved):

Issue #77, opened 2023-01-27
`Upgradable::up_stage_code` parameter serialization


mooori:
`Upgradable::up_stage_code` uses borsh parameter serialization:

https://github.com/aurora-is-near/near-plugins/blob/c043add4c2a0810872c4326a55c5162bccd3f4ee/near-plugins-derive/src/upgradable.rs#L33

However, this isn’t documented:

https://github.com/aurora-is-near/near-plugins/blob/1cf0aa61e38d3b02803642983795c927598375ba/near-plugins/src/upgradable.rs#L56-L58

So most likely users will assume default json parameter serialization and then wonder why deserialization fails when calling `up_stage_code` in a transaction.

---

Either we should document this or change `up_stage_code` to use json serialization. If I remember correctly, this is the only method provided by near-plugins which does _not_ use json parameter serialization. Even if less efficient, perhaps we should switch to json here for a consistent and intuitive API?

cc @birchmd

mooori:
Also the borsh serialization of `up_staged_code` isn’t documented:

https://github.com/aurora-is-near/near-plugins/blob/1cf0aa61e38d3b02803642983795c927598375ba/near-plugins/src/upgradable.rs#L60-L61

https://github.com/aurora-is-near/near-plugins/blob/c043add4c2a0810872c4326a55c5162bccd3f4ee/near-plugins-derive/src/upgradable.rs#L41-L44

Since it’s related to `up_stage_code`, I think we can include it in this issue and handle it in the same way (document or switch to json).

birchmd:
Good catch @mooori . I think it makes sense to change the serialization to use JSON (bytes encoded as base64) to be consistent with everything else. If the efficiency is really important to some use-case they can always override the default implementation correct?

mooori:
It surfaced in integration tests and surprised me there.

If someone implements the `Upgradable` trait themselves, they can choose another serialization protocol. Using `#[derive(Upgradable)]` and overriding just a single trait method might be tricky or even impossible.

Another simple way to use borsh instead of json would be adding a wrapper like this to the contract:

```rust
pub fn borsh_up_stage_code(&mut self, #[serializer(borsh)] code: Vec<u8>) {
    self.up_stage_code(code);
}
```

Given these options I think for now it’s fine to switch to json. If efficiency of these methods becomes a frequent issue, we could let users configure the serialization protocol via a macro attribute:

```rust
#[derive(Upgradable)]
[...]

PR #117, opened 2023-11-01
feat(upgradable)!: add `hash` parameter to `up_deploy_code`

mooori:
Adds the parameter `hash: Option<String>` to the function `up_deploy_code` of the `Upgradable` plugin.

If `hash` is `h`, the deployment succeeds only if `h` equals the hash of the currently staged code. In particular, `h` must correspond to the base64 encoded sha256 hash of the deployed code. Given the bytes of the deployed code, `h` can be computed using near-sdk-rs. See for instance the function [convert_code_to_deploy_hash](https://github.com/aurora-is-near/near-plugins/blob/974c3412551c26e029c2abecedfff5c5a954121a/near-plugins-derive/tests/upgradable.rs#L215-L218) used in tests.

**Motivation**: It helps facilitate some workflows, for example [DAO upgrade approval](https://github.com/aurora-is-near/near-plugins/issues/77#issuecomment-1772832387).

### BREAKING CHANGE

This is a breaking change since the signature of `up_deploy_code` is changed.

```rust
let promise = up_deploy_code(hash, function_call_args);
```


birchmd (commented):
The code looks good to me, but I have some questions on the design.

If this is a breaking change then I do not see a reason to make the hash optional. I assume the DAO should always sign over a hash to make sure the right code is deployed, so having the `None` escape hatch seems like it creates unnecessary risk.

We could avoid the breaking change by having the hash verification be a separate function in the trait (e.g. `up_verified_deploy_code`), and again in that function the hash would not be optional. In this version the escape hatch of the old function is still available, but we could mark it as deprecated and remove it in a future release instead of making the breaking change now.

birchmd:
>  not due unsafety of the version without `hash`. If that’s the case, we might not necessarily have to deprecate and remove it.

When I say "unsafe" I mean with respect to the DAO-controlled workflow. Essentially the DAO is not signing over the actual bytes being deployed (because the transaction is too large), so instead they are signing over the hash of the bytes as part of the transaction that actually triggers the upgrade. If that hash were missing (either because `None` is an allowed value, or they use the old function without any hash) then they may not be approving the bytes to deploy that they intended to.

In general the flow is not unsafe if the role which triggers the upgrade trusts the role which stages the bytes to deploy. If these roles do not trust each other then the hash is needed to ensure they agree on what bytes are actually being deployed.

Whether or not we keep the old version of the function (without any hash) depends on if we think the case of the roles trusting each other is common enough that including the hash will be unnecessary for many users. Personally I think it is better to always include the hash, even if the roles trust each other, because a redundant check only decreases the probability of making a mistake during the upgrade.

karim-en:
Replaced Option<hash> with hash
aeff8256d4e98ef402d3ff6b6dca5785636b86b0

Garrybrayn (commented):
<>

olga24912 (approved):

karim-en (approved):

PR #118, opened 2023-11-02
build!: update Rust version and `near-*` dependencies

mooori:
With the current minimum `rust-version` of 1.64.0 tests cannot be built anymore, e.g. see [this CI failure](https://github.com/aurora-is-near/near-plugins/actions/runs/6719053645/job/18259922214?pr=117). Hence this PR increseases `rust-version` to ~1.70.0~ 1.69.0 and updates `near-workspaces`. An update of both `rust-version` and `near-workspaces` is required to make tests build and pass again, hence this PR combines both.

### Why ~1.70.0~ 1.69.0

The [latest version](https://crates.io/crates/near-workspaces/0.9.0) of `near-workspaces` requires at least Rust 1.69.0.

### BREAKING CHANGES

The minimum supported Rust version is increased from 1.64.0 to ~1.70.0~ 1.69.0.

### `near-workspaces`

- Users of `near-plugins` should not be affected by this update since it is a `dev-dependency` only used in tests.
- The crate was renamed from `workspaces` to `near-workspaces` and many lines in the diff are due to this change.

### `near-sdk`

Updating from `4.1.0` to `4.1.1` since Near dependencies are already touched in this PR.

### Outlook

Some other dependencies had major versions bumps and we should update them as well. It might require some refactors, so I would suggest to that in separate PR(s). I will open a tracking issue shortly.


birchmd (approved):

karim-en (changes requested):
Could you please verify that it is still possible to build contracts with plugins using rust 1.69, because currently there is an issue with parsing the contracts by [nearblocks](https://nearblocks.io/) if the rust version is 1.70 and higher.
Check this: https://github.com/aurora-is-near/rainbow-token-connector/pull/234/files

mooori:
@karim-en thank you for bringing this up. As you were running into the same issue, I tried to figure out the underlying problem. See [this issue](https://github.com/near/near-workspaces-rs/issues/336).

To verify `near-plugins` works with Rust 1.69.0, I will try to use 1.69.0 in the contracts in `near-plugins-derive/tests/contracts` and run the tests with Rust 1.70.0. (Similar to what you did with `rainbow-token-connector`).

mooori:
Turns out that the failure to build `near-workspaces` with Rust 1.69.0 is a [known issue](https://github.com/near/near-workspaces-rs/issues/336#issuecomment-1792415901). It is recommended to downgrade packages manually, which is done in [`scripts/fix-dependencies`](https://github.com/aurora-is-near/near-plugins/blob/8d95f01004a06b80b96b680540861a695febe92c/scripts/fix_dependencies.sh). It is likely to require some maintenance as these dependencies are upgraded - please have a look at the comment in the script. Still, it seems to be the only option I can make work in CI since all my other attempts (see the CI failures in #119) have failed.

As a result, the MSRV can be lowered to 1.69.0, which is used to compile test contracts and in CI.

birchmd (approved):
This explicit downgrading of dependencies is a little awkward so if possible I think it would be nice to get the MSRV up to 1.70 (once this is possible for the bridge). But for now this looks like a good work-around.

karim-en:
> This explicit downgrading of dependencies is a little awkward so if possible I think it would be nice to get the MSRV up to 1.70 (once this is possible for the bridge). But for now this looks like a good work-around.

Actually, the contracts that are built with 1.70 work correctly because the protocol itself has already been upgraded to support those contracts. 
So to be able to upgrade I think we need to contribute to this repo https://github.com/NEARFoundation/near-contract-parser/tree/main, and then ask near https://nearblocks.io/ owners to update this crate.

mooori:
I've opened #121 for the MSRV upgrade.

karim-en (approved):

mrLSD (approved):

PR #122, opened 2023-11-14
build: only use `borsh` re-exported by `near-sdk`

mooori:
`near-plugins-derive` was the only workspace member that depended on `borsh`. However, it isn’t actually used. Instead `borsh` re-exported by `near-sdk` is used, which is considered [best practice](https://docs.near.org/sdk/rust/best-practices#reuse-crates-from-near-sdk). Removing the `borsh` dependency might also prevent using it accidentally (instead of `borsh` re-exported by `near-sdk`) in the future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ near-plugins = { path = "near-plugins" }
 near-plugins-derive = { path = "near-plugins-derive" }
 serde = "1"
 anyhow = "1.0"
-borsh = "0.9"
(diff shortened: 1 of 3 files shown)


birchmd (approved):
Aurora Engine uses [`udeps` to catch unused Cargo dependencies](https://crates.io/crates/cargo-udeps). Maybe it makes sense to add it to CI here too.

karim-en (approved):

mrLSD (approved):

PR #128, opened 2024-01-18
build: add workarounds required to build tests

mooori:
## Override `parity-secp256k1`

Building tests may currently fail since a dependency of `near-sdk` v4.1.1 (most recent release) has been yanked. Providing a patch is the [recommended workaround](https://github.com/near/near-sdk-rs/issues/1119#issuecomment-1889323112).

A new version of `near-sdk` is planned to be released soon. Then this patch shouldn’t be required anymore ([ref.](https://github.com/near/near-sdk-rs/issues/1119#issuecomment-1889323112)). 

## Update `scripts/fix-dependencies`

By now more recursive dependencies need to be downgraded manually for tests to build with MSRV 1.69.0. See [this comment](https://github.com/aurora-is-near/near-plugins/pull/118#issuecomment-1794576809) for details.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,3 +35,8 @@ proc-macro2 = "1.0"
 quote = "1.0.9"
 syn = { version = "1.0.69", features = ["full"] }
 proc-macro-crate = "0.1.5"
+
+# Required to build tests with near-sdk v4.1.1, see #128.
+# TODO(#125): Remove after upgrading to near-sdk v5.
+[patch.crates-io]
+parity-secp256k1 = {git = "https://github.com/paritytech/rust-secp256k1", tag = "parity-secp256k1-v0.7.0"}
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -35,3 +35,8 @@ proc-macro2 = "1.0"
 quote = "1.0.9"
 syn = { version = "1.0.69", features = ["full"] }
 proc-macro-crate = "0.1.5"
+
+# Required to build tests with near-sdk v4.1.1, see
+# TODO(): Remove after upgrading to near-sdk v5.
+[patch.crates-io]
+parity-secp256k1 = {git = "https://github.com/paritytech/rust-secp256k1"}
```
aleksuss, line 42: I would recommend adding a hash commit because, in such a way, you download the latest code from the master branch which potentially could be broken.

mooori: Good point! I’ve added a `tag` corresponding to the last release of the `parity-secp256k1` crate.

Would you mind pointing that out [here](https://github.com/near/near-sdk-rs/issues/1126) as well? The workaround recommended to the community is currently missing a `tag` (or anything else that pins the downloaded code).

birchmd (approved):

PR #130, opened 2024-01-19
Update toolchain and manifest to Rust 1.70.0

mikedotexe:
I noticed this error when opening the repo:

Screenshot 2024-01-19 at 8 37 40 AM

So updated `Cargo.toml` and `rust-toolchain` in the project root from `1.69.0` » `1.70.0`

This repo looks really amazing, btw, bravo!


mooori:
Thank you for bringing this up! We have [workarounds](https://github.com/aurora-is-near/near-plugins/blob/542601c3dae8229e2caf5d723cc10b6d83a01f42/scripts/fix_dependencies.sh) for building with 1.69 and use them in CI, but missed to highlight them to users.

We cannot yet bump the MSRV as some users of this crate still need to use Rust 1.69.0. To prevent issues like this, I’ve opened #131. After it lands, user with a toolchain >= 1.70.0 should not hit this issue anymore, as the override is removed from `rust-toolchain` ([ref](https://rust-lang.github.io/rustup/overrides.html)).

PR #134, opened 2024-05-27
feat: bump near-sdk to 5.1.0

mitinarseny:


near-plugins-derive/src/access_controllable.rs
```diff
@@ -49,14 +50,15 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
         #input
 
         #[derive(::near_sdk::borsh::BorshDeserialize, ::near_sdk::borsh::BorshSerialize)]
+        #[borsh(crate = "near_sdk::borsh")]
         struct #acl_type {
             /// Stores permissions per account.
-            permissions: ::near_sdk::store::UnorderedMap<
+            permissions: ::near_sdk::store::LookupMap<
```
birchmd, line 56: What is the motivation for the change from `UnorderedMap` to `LookupMap`? I think it will break the ACL plugin because it requires being able to iterate over the keys of the map for introspective functions.

near-plugins-derive/tests/upgradable.rs
```diff
@@ -599,10 +599,10 @@ async fn test_deploy_code_in_batch_transaction_pitfall() -> anyhow::Result<()> {
         .args_json(json!({ "function_call_args": FunctionCallArgs {
         function_name: "migrate_with_failure".to_string(),
         arguments: Vec::new(),
-        amount: 0,
-        gas: Gas::ONE_TERA,
+        amount: NearToken::from_yoctonear(0),
+        gas: Gas::from_tgas(1),
     } }))
-        .gas(near_workspaces::types::Gas::from_tgas(200));
+        .gas(near_workspaces::types::Gas::from_tgas(201));
```
birchmd, line 605: Is bumping it up from 200 to 201 necessary?

near-plugins-derive/src/access_controllable.rs
aleksuss: The point is that `near_sdk::store::UnorderedMap` is already deprecated in the near_sdk 5.1.0. But of course, if we need to iterate over the permissions, then we probably should use `near_sdk::collections::UnorderedMap` here.

near-plugins-derive/tests/upgradable.rs
aleksuss: No. It's not. Will be reverted in the [PR](https://github.com/mitinarseny/near-plugins/pull/2).

.gitignore
```diff
@@ -1,5 +1,4 @@
 /target
-Cargo.lock
```
aleksuss, line 2: Why?

near-plugins-derive/src/access_controllable.rs
birchmd: We were on `collections` originally and migrated to `store` because it looked like `collections` itself might become deprecated. [That prior PR](https://github.com/Near-One/near-plugins/pull/37) points out that the representation of the `UnorderedMap` is different between the two modules so it would be a breaking change to return to `collections` and potentially a difficult migration for the projects using a previous version of near-plugins.

aleksuss: As I understand, they don't recommend using [iterable collections](https://github.com/near/near-sdk-rs/blob/near-sdk-v5.1.0/near-sdk/src/store/unordered_map/mod.rs#L89) at all.

.gitignore
birchmd: https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

The motivation could be to make the CI more stable. Though it is worth thinking about if we should add `cargo update` to one of the CI flows so that we can know if a newer version of a dependency causes the library to break.

near-plugins-derive/src/access_controllable.rs
birchmd: I don't think the iteration performance issue is really a big deal for our use case because these introspective methods that iterate over the map are meant primarily to be view functions I think. So the users are not paying for the gas spent.

mitinarseny: I think, the main concern here is that `UnorderedSet` (and `UnorderedMap`) [stores](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/unordered_map/mod.rs#L102) and deserializes the [entire](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/free_list/mod.rs#L35) list of keys every time users interacts with a contract.

I assume that a typical use-case for `AccessControllable` would be to check permissions for some particular user. So, if a user really needs to iterate over all account IDs, IMHO it's better to use off-chain indexers that would provide the historical info on that.

I agree on a point that using `LookupSet` (or `LookupMap`) would introduce a breaking change. So, maybe we can warn developers by bumping the major version for `near-plugins` but still make it more gas-efficient?

aleksuss: > I think, the main concern here is that UnorderedSet (and UnorderedMap) [stores](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/unordered_map/mod.rs#L102) and deserializes the [entire](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/free_list/mod.rs#L35) list of keys every time a user interacts with a contract.

No. The behaviour of the `UnorderedMap` is the same as `LookupMap`, but the first one [supports iteration](https://docs.rs/near-sdk/5.1.0/near_sdk/store/unordered_map/struct.UnorderedMap.html). You described the behaviour of the `std::collections::UnorderedMap`.
 
> So, maybe we can warn developers and bump the major version for near-plugins but still make it more gas-efficient?

Having a major version 0 [means](https://semver.org/#spec-item-4) that the API is unstable and could be changed at any time. By increasing the minor version, we make a notice that we bring breaking change. 

The main concern here is that after replacing the `UnorderedMap` with the`LookupMap` we will have to do a state migration of the contract if we want to update near-plugins to 0.3.0. (I believe that UnorderedMap and LookupMap have different byte representations after serialization).

mitinarseny: > No. The behaviour of the UnorderedMap is the same as LookupMap, but the first one [supports iteration](https://docs.rs/near-sdk/5.1.0/near_sdk/store/unordered_map/struct.UnorderedMap.html). You described the behaviour of the std::collections::UnorderedMap.

TLDR: it's not as bad as I thought initially, but can still be optimized out and save gas for users

I'm talking about `UnorderedMap` and `UnorderedSet` from `near_sdk::store`. Support for iteration comes with a cost of having additional field [`keys: FreeList<K>`](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/unordered_map/mod.rs#L102), which internally has a field [`elements: Vector<Slot<T>>`](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/free_list/mod.rs#L35). After reading source code for [`Vector`](https://docs.rs/near-sdk/latest/near_sdk/store/vec/struct.Vector.html) I agree that it doesn't deserialize the entire list from the storage and load elements lazily from underlying  [`values: IndexMap<T>`](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/vec/mod.rs#L129) only when it got to the point when iteration is actually happening.

On the other hand, it still does a bunch of unnecessary calculations and deserialization. An example would be a `Drop` implementation for `IndexMap<T>` which calls [`IndexMap::flush`](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-sdk/src/store/index_map.rs#L50-L78) internally. Yes, it iterates only through the cached items and writes back to the storage only modified elements (an element is considered as modified if at least one of following methods were called: [`Vector::set`](https://docs.rs/near-sdk/latest/near_sdk/store/vec/struct.Vector.html#method.set), [`Vector::get_mut`](https://docs.rs/near-sdk/latest/near_sdk/store/vec/struct.Vector.html#method.get_mut) or [`Vector::iter_mut`](https://docs.rs/near-sdk/latest/near_sdk/store/vec/struct.Vector.html#method.iter_mut)). 

Thus, it means that if you use `#[near_sdk::near]` attribute on your contract and all methods that accept `&mut self`, then you're potentially doing unnecessary things as the storage of a contract would be [serialized and written back to the storage](https://docs.near.org/sdk/rust/contract-interface/contract-mutability#mutable-functions) if you call methods that take `&mut self`).

> The main concern here is that after replacing the UnorderedMap with theLookupMap we will have to do a state migration of the contract if we want to update near-plugins to 0.3.0. 

Basically, we have two options: either leave it as it was before or introduce a breaking change and optimize gas consumption for future users. I don't have an exact number on how much would it optimize gas usage, so maybe you're right and it doesn't worth it. 

On the other hand, if this PR gets merged then it would be a breaking change anyway since we're now using a renewed API of `near_sdk`, and developers would still have to handle those API changes in their code as well.

.gitignore
mitinarseny: Thank you for the comment, removed

near-plugins-derive/src/access_controllable.rs
birchmd: >  it's better to use off-chain indexers

View functions are much more convenient than running an indexer, which the barrier to accessing the ACL significantly lower. I agree that if a project was going for maximal efficiency an indexer is better, but I think ease of use is more important than 5%-10% less gas used (I'm making up numbers here, but I expect that the overhead for normal operation is in that range).

Since the ability to view the ACL exists today, I think it would be our responsibility to provide an indexer implementation and documentation for how to migrate from a workflow that uses view functions to one that uses the indexer. That sounds like much more work than we want to take on for a small reduction in gas usage.

> it would be a breaking change anyway

Echoing @aleksuss , I think it is important to distinguish API changes from changes requiring state migration. Both are breaking, but the latter is much more work for a project to deal with. We would also need to provide the state migration code and documentation for how to use it, so it more work for us too.

Overall, I strongly recommend that we keep using `store::UnorderedMap` for now even though it is deprecated. We could open a discussion with the bridge team (I think they are the primary users of this library) in the meantime about if they are worried about the deprecation and/or the gas cost, and if they are open to an indexer solution instead of a view function solution.

mooori (commented):
To follow the v5 docs of near-sdk, I think it would be good to use `#[near]` [instead of](https://github.com/Near-One/near-plugins/blob/2d7b4784fc594059f88c357c712fea876fff92cc/near-plugins-derive/tests/contracts/ownable/src/lib.rs#L12) `#[near_bindgen]` in the test/example contracts.

.github/workflows/test.yml
```diff
@@ -17,16 +17,13 @@ jobs:
       - name: Install toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.69.0 # MSRV
+          toolchain: 1.74.0 # MSRV
```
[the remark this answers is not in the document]
mitinarseny: Hey @mooori,

Unfortunately, `near_sdk 5.1` [requires](https://github.com/near/near-sdk-rs?tab=readme-ov-file#msrv) MSRV of `1.72`. @aleksuss what is the reason for bumping it even more up to `1.74`?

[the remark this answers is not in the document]
aleksuss: The `clap_builder 4.5.2` [requires](https://github.com/mitinarseny/near-plugins/actions/runs/9255815271/job/25460403459#step:6:407) MSRV 1.74.0.

near-plugins-derive/tests/upgradable.rs
mitinarseny: Unfortunately, gas consumption went up a bit. It's likely due to a bit heavier logic in updated dependencies such as borsh/near-sdk/etc...

mooori:
Strangely check `Rust Contracts / Tests (pull_requests)` was skipped after approving workflows to run, even though relevant files where changed since the last run. It didn't tell why so I just started it manually through Github's UI.

near-plugins-derive/src/access_controllable.rs
```diff
@@ -53,13 +53,16 @@ pub fn access_controllable(attrs: TokenStream, item: TokenStream) -> TokenStream
         #[borsh(crate = "near_sdk::borsh")]
         struct #acl_type {
             /// Stores permissions per account.
+            #[allow(deprecated)]
```
mooori, line 56: I think it would be helpful to add a short comment on `#acl_type` that all `store` types which make sense here have been deprecated. Otherwise future readers might wonder about this.

It seems unfortunate that all `store` types which are iterable have been deprecated by `near-sdk`. Perhaps [TreeMap](https://docs.rs/near-sdk/latest/near_sdk/store/index.html) will be stabilized, but we don't need ordered keys here and shouldn't pay the cost for them.

mitinarseny: Done

mooori (approved):
Looks great, thank you!

mooori:
I missed [542009b](https://github.com/Near-One/near-plugins/pull/134/commits/542009b3a373d888dd7b7bf90051b6ff3f18cb1d) was pushed while I was reviewing. Will have to take a look at that before enabling merge.

mooori:
@mitinarseny what is the motivation for adding [542009b](https://github.com/Near-One/near-plugins/pull/134/commits/542009b3a373d888dd7b7bf90051b6ff3f18cb1d) to this PR?

mitinarseny:
I've just tried to use `near-plugins` in another project and it turns out that it doesn't let you use `#[derive(AccessControlRole)]`, `#[access_control]`, etc... without importing indirect trait dependencies into current scope:
```rust
use near_plugins::{access_control, AccessControlRole, AccessControllable, Ownable, Pausable};

#[derive(AccessControlRole, Clone, Copy)]
pub enum Role {
    // ...
}

#[access_control(role_type(Role))]
pub struct Contract {
    // ...
}
```

But I agree that it's not directly connected to "bumping near-sdk version".
@mooori Do you think it would be better to fix it in a separate PR?

karim-en:
> PS: I might need to rebase the whole branch to have my commits signed... Do you mind if I do it now and re-request your review? Sorry for that, didn't pay attention in the first place

I want to avoid the force push and re-review, so I've disabled this requirement.

karim-en (approved):

mitinarseny:
Thank you so much! @mooori @karim-en 🙏🏻
Waiting for merge from someone with write access

PS: Already set up everything related to commit signing, so wouldn't mess up in the next PR😅

PR #137, opened 2024-07-11
deps: bump near-sdk to 5.2

aleksuss:


Cargo.toml
```diff
@@ -13,7 +13,7 @@ authors = ["Aurora Labs <hello@aurora.dev>"]
 # An update of the MSRV requires updating:
 # - `rust-toolchain` files in `near-plugins-derive/tests/contracts/**`
 # - the toolchain installed in CI via the `toolchain` parameter of `actions-rs/toolchain@v1`
-rust-version = "1.74.0"
+rust-version = "1.76.0"
```
mooori, line 16: Is the MSRV bump required due to dependency updates?

Cargo.toml
```diff
@@ -23,14 +23,14 @@ keywords = ["near", "smart contract", "plugin"]
 
 [workspace.dependencies]
 bitflags = "1.3"
-near-sdk = "5.1.0"
+near-sdk = "5.2"
```
mooori, line 26: This could allow getting rid of deprecated `UnorderedMaps`, see #138.

mooori (approved):
LGTM.

If bumping the MSRV is not strictly required, then it would be better to not change it to avoid issues for users of the plugins. Some may not yet be ready to bump their project's MSRV.

Cargo.toml
aleksuss: Unfortunately yes. The `near-workspaces 0.11` required 1.76.0.

mooori: Heads up for the Bridge team that the MSRV is bumped here, cc @karim-en.

PR #139, opened 2024-07-15
feat: use new `IterableMap` and `IterableSet` collections

aleksuss:
TODO: Check backward compatibility.


Cargo.toml
```diff
@@ -36,3 +36,8 @@ proc-macro2 = "1.0"
 quote = "1.0.9"
 syn = { version = "1.0.69", features = ["full"] }
 proc-macro-crate = "0.1.5"
+
+[workspace.lints.clippy]
+all = "deny"
```
Garrybrayn, line 41: garrybrayn

Garrybrayn: garrybrayn

Cargo.toml
```diff
@@ -36,3 +36,8 @@ proc-macro2 = "1.0"
...
+nursery = "deny"
+pedantic = "deny"
```
Garrybrayn, line 43: garrybrayn

Garrybrayn (changes requested):

karim-en:
@aleksuss  Looks like the changes are backward compatible. We have added a test [here](https://github.com/Near-One/near-plugins/pull/141), please merge with master to check it out.

aleksuss:
> Looks like the changes are backward compatible.

@karim-en  But the test says the opposite.

karim-en:
> > Looks like the changes are backward compatible.
> 
> @karim-en  But the test says the opposite.
Sorry it was a typo. I wanted to say "aren't backward compatible"

aleksuss:
I thought to add these changes to the 0.3.0 version, but as I can see, this version has already been released.

karim-en:
> I thought to add these changes to the 0.3.0 version, but as I can see, this version has already been released.

It is dangerous to do a release like this because the plugin's users would like to upgrade to the latest version and miss the plugins being upgradable and needing a migration.
What I can propose is adding a non-default feature flag for these changes.

aleksuss:
Changing a minor version if the major is 0 implies the existence of breaking changes by semver rules. As for me, I don't like a feature flag. I would postpone these changes to the next release though.

karim-en:
> Changing a minor version if the major is 0 implies the existence of breaking changes by semver rules. As for me, I don't like a feature flag. I would postpone these changes to the next release though.

It is storage breaking changes, we can't allow releasing things like this without proper storage migration implementation.
The breaking changes are acceptable just on the code or API level.

PR #144, opened 2024-09-10
fix: make view methods pausable

karim-en:
- Skip role verifications if the pausable method doesn't contain except roles to able to call view methods.
- Fix the `near-sdk` version to avoid auto upgrade that breaks CI and build due to dependencies conflicts.

### DIFF
--- a/near-plugins-derive/tests/pausable.rs
+++ b/near-plugins-derive/tests/pausable.rs
@@ -342,6 +342,18 @@ async fn test_not_paused_with_different_key() -> anyhow::Result<()> {
 }
 
 #[tokio::test]
+async fn test_view_call_of_pausable_method() -> anyhow::Result<()> {
+    let setup = Setup::new().await?;
+    let res = setup
+        .call_counter_modifier(&setup.unauth_account, "increase_1")
+        .await?;
+    assert_success_with_unit_return(res);
+    assert_eq!(setup.get_counter().await?, 1);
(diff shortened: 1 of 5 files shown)


frolvanya (approved):
There're small nitpicks like `args.roles.len() == 0` -> `!args.is_empty()` or `let err = format!("{}", err);` -> `let err = err.to_string();`, but they're not that important, so can be ignored

near-plugins-derive/tests/pausable.rs
```diff
@@ -341,6 +347,25 @@ async fn test_not_paused_with_different_key() -> anyhow::Result<()> {
...
+        .call_counter_modifier(&setup.unauth_account, "increase_1")
+        .await?;
+    assert_success_with_unit_return(res);
+    assert_eq!(setup.get_counter().await?, 1);
+
+    let res = setup
+        .pausable_contract
+        .pa_pause_feature(&setup.pause_manager, "get_counter")
+        .await?;
+    assert_success_with(res, true);
```
olga24912, line 363: We should also call the `get_counter` as a view method to ensure that we are able to do so.

olga24912 (approved):

near-plugins-derive/tests/pausable.rs
karim-en: We already do a view call line below

PR #153, opened 2025-02-24
feat: bump CI to Rust 1.83

r-near:


karim-en:
@r-near to fix the CI I think we need to use the `cargo_near_build` here https://github.com/Near-One/near-plugins/blob/7b47dbe1275259a9e3969b8ea281531e6e01048d/near-plugins-derive/tests/common/repo.rs#L38 like we did it here https://github.com/Near-One/omni-bridge/blob/f1ad7015414e6cff49d7ba793c13742eb02ff380/near/omni-tests/src/helpers.rs#L20

PR #156, opened 2025-02-25
feat: Separate pause and unpause roles in Pausable plugin

### LINKED ISSUE #124: `Pausable`: add pause & unpause roles
Currently, the features can be paused and unpaused by the one-role `PauseManager`, which makes it impossible to assign just a `Pause` role without `Unpause`.
There is a workaround for this by creating an account with limited access to the pause method and assigning the `PauseManager` role to it. Unfortunately, this complicates the access management, so it would be good if it would be possible to add additional roles like in the `Upgradable`.

### LINKED ISSUE #151: Split pause and unpause roles
Currently, it is not possible to set different roles for pause and unpause, there is just one `PasueManager` role, and we need to fix that.

r-near:
Replaces the single `manager_roles` attribute with separate `pause_roles` and `unpause_roles` attributes in the Pausable plugin. This enables assigning different permissions for pausing and unpausing features.

## Implementation

- Modified `pausable.rs` trait and implementation to support separate role specifications
- Updated test contracts and tests
- Added migration documentation and CHANGELOG entry

## Usage

```rust
// Before
#[pausable(manager_roles(Role::PauseManager))]

// After
#[pausable(
    pause_roles(Role::PauseManager),
    unpause_roles(Role::UnpauseManager)
)]
```

## Impact

Breaking change that requires code updates when upgrading. This enables more granular security policies, particularly in scenarios where emergency pausing should be available to more accounts than unpausing.

- Closes #151
- Closes #124

### DIFF
--- a/docs/migrations/pausable-separate-roles.md
+++ b/docs/migrations/pausable-separate-roles.md
@@ -1,0 +1,114 @@
+# Migrating to Separate Pause/Unpause Roles
+
+This guide explains how to migrate your code to use the new `pause_roles` and `unpause_roles` attributes instead of the consolidated `manager_roles` attribute in the Pausable plugin.
+
+## Changes Required
+
+### Before
+
+Previously, you would define permissions for both pausing and unpausing using a single attribute:
+
+```rust
+#[pausable(manager_roles(Role::PauseManager))]
+struct Contract {
+    // Contract fields
+}
+```
+
+This meant that any account with the `PauseManager` role could both pause and unpause features.
+
+### After
+
+Now, you need to specify permissions for pausing and unpausing separately:
+
+```rust
+#[pausable(
+    pause_roles(Role::PauseManager),
+    unpause_roles(Role::UnpauseManager)
+)]
+struct Contract {
+    // Contract fields
+}
+```
+
+With this change, you can:
+- Grant an account only the ability to pause features (emergency response)
+- Grant a different account only the ability to unpause features (recovery process)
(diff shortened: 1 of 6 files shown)


frolvanya (approved):

karim-en (approved):

kiseln (approved):

docs/migrations/pausable-separate-roles.md
```diff
@@ -0,0 +1,114 @@
...
+## Step-by-Step Migration
+
+1. **Update your Role enum** to include separate roles for pausing and unpausing, if desired:
+
+   ```rust
+   #[derive(AccessControlRole, Deserialize, Serialize, Copy, Clone)]
+   #[serde(crate = "near_sdk::serde")]
+   pub enum Role {
+       // Previous role that could both pause and unpause
+       // PauseManager,
```
mitinarseny, line 48: You should never remove a variant or add one more in the middle in enums with `#[derive(AccessControlRole)]`, since it [uses](https://github.com/Near-One/near-plugins/blob/e0cb09455ebb16c346d86c40aac3f926080195ae/near-plugins-derive/src/access_control_role.rs#L204-L209) bitflags to store permissions for each `AccountId`.
Please, fix the migration guide.

r-near: @karim-en PR here: https://github.com/Near-One/near-plugins/pull/163

PR #157, opened 2025-02-26
chore: deploy test contracts with cargo-near

karim-en:


near-plugins-derive/tests/common/repo.rs
```diff
@@ -23,45 +9,18 @@ pub fn require_success(output: Output) -> Result<(), anyhow::Error> {
...
-        ])
-        .output()
-        .await?;
+    let path = project_path.join("Cargo.toml");
+    let artifact = cargo_near_build::build(cargo_near_build::BuildOpts {
+        manifest_path: Some(
+            cargo_near_build::camino::Utf8PathBuf::from_str(path.to_str().unwrap())
+                .expect("camino PathBuf from str"),
+        ),
+        no_abi: true, // TODO remove this flag when we fix ABI generation
```
frolvanya, line 19: Are there many errors right now? I thought that `near-plugins` were fixed to support abi

frolvanya (approved):

near-plugins-derive/tests/common/repo.rs
karim-en: Looks like not all issues are fixed.
@r-near can you please verify and fix that?

frolvanya: probably there're abi issue during the compilation, you need to use `#[near(serializers = [borsh, json])]` to fix it

PR #196, opened 2026-06-05
fix: detect near_bindgen-generated functions on rustc >= 1.87

ricky-sb:
## Problem

On any `rustc >= 1.87`, every near-plugins attribute macro (`#[only]`, `#[pause]`, `#[if_paused]`, `#[access_control_any]`, ...) fails to compile a contract for the `wasm32-unknown-unknown` target with:

```
error[E0424]: expected value, found module `self`
   = note: this error originates in the attribute macro `only`
```

Minimal repro (validated): near-sdk `5.7.1` + near-plugins `0.5.1`, a single `#[only(owner)]` method, `cargo +1.93.0 build --target wasm32-unknown-unknown --release`.

```rust
#[near(contract_state)]
#[derive(Ownable, Default)]
pub struct Counter { counter: u64 }

#[near]
impl Counter {
    #[only(owner)]
    pub fn protected(&mut self) { self.counter += 1; }
}
```

Reproduces on 1.87 through 1.96; compiles fine on 1.86.

## Root cause

`near_bindgen` generates `extern "C"` wrapper / marshall functions for each contract method, annotated with `#[cfg(target_arch = "wasm32")]`. `is_near_bindgen_wrapped_or_marshall` (in `near-plugins-derive/src/utils.rs`) detected these generated functions by string-matching that `#[cfg(target_arch = "wasm32")]` attribute, so the plugin macro would skip injecting its `self.owner_is()` / pause / ACL checks into them.

Since **rustc 1.87**, `#[cfg]` attributes are evaluated and stripped *before* attribute proc-macros expand. By the time the plugin macro re-expands on the generated wrapper, the `#[cfg(target_arch = "wasm32")]` marker is already gone, so detection returns `false`. The macro then injects `self.owner_is(...)` into the generated wrapper — which has no `self` receiver — producing `E0424`.

This is independent of the near-sdk version; it depends only on the compiler. It went unnoticed because cargo-near previously capped contract builds at rustc 1.86; the recent 2.12 / PV-84 toolchain unlock (rustc 1.93) removed that cap and unmasked it. It was reported by a user building a contract with rustc 1.93.

## Fix

Add a receiver guard at the top of `is_near_bindgen_wrapped_or_marshall`: the generated wrapper / marshall functions never take a `self` receiver, while the plugin macros are only ever meaningful on methods (which do). So any function without a `self` receiver can be safely treated as generated and skipped.

```rust
if item.sig.receiver().is_none() {
    return true;
}
```

This does not rely on `#[cfg]` surviving, so it works on every toolchain. The original cfg-based string match is kept below it as a belt-and-braces fallback for older compilers. Because all four plugin macros call this one helper, the single guard fixes all of them.

## Verification

- `cargo +1.93.0 build --target wasm32-unknown-unknown --release` on the repro contract: **fails with E0424 before this change, succeeds after**.
- `cargo clippy --all --all-features --all-targets -- -D warnings` and `cargo fmt --all -- --check` on MSRV 1.86.0: clean.
- `cargo test --all --all-features` builds all example contracts under `tests/contracts/**` successfully on 1.86.0 (no behavioral change on the existing toolchain; the guard only short-circuits earlier than the existing heuristic).

## Notes

- No test was added to CI: the suite pins every crate and example contract to MSRV 1.86.0, where the bug cannot reproduce. Exercising it would require introducing a `>= 1.87` toolchain into CI, which is a larger CI change; the repro is documented above instead. Happy to add a CI job on a newer toolchain in a follow-up if you'd prefer.
- Diff is intentionally minimal (8 lines, one file). The changelog / version bump (0.5.2) should be picked up automatically by release-plz from the `fix:` commit.

Maintainer review + a 0.5.2 release would unblock contract authors on the new toolchain. cc @frol (I believe @vsavchyn-dev is out until ~June 12).


r-near:
Superseded by #197 (recreated from a branch in this repo under the correct account). Closing this one.

PR #198, opened 2026-06-15
fix: ci publishing error due to versioned dev-dep in `near-plugins-derive`

vsavchyn-dev:
Right now `release-plz` workflow for `near-plugins` and `near-plugins-derive` fails as `cargo-publish` cannot find `0.5.2` version of `near-plugins` for dev dep of `near-plugins-derive`: https://github.com/near/near-plugins/actions/runs/27223151434/job/81470121038#step:4:339

This PR aims to fix this issue in CI and have a follow-up to publish previously mentioned crates as `0.5.2`.

NOTE:
- The `release-plz` pr might increase crates' version to 0.5.3 - we need to set it back and release it as 0.5.2


frolvanya:
Don't you need to revert this version as well: https://github.com/near/near-plugins/blob/4fcede61bac37cddd7ff87091b48b48b1551ff31/near-plugins/Cargo.toml#L20?

near-plugins-derive/Cargo.toml
```diff
@@ -23,7 +23,7 @@ proc-macro-crate.workspace = true
 
 [dev-dependencies]
 anyhow.workspace = true
-near-plugins = { path = "../near-plugins/", version = "0.5.2" }
+near-plugins = { path = "../near-plugins/" }
```
frolvanya, line 26: Any particular reason for unpinning version?
I'd just revert #188

```suggestion
near-plugins = { path = "../near-plugins/", version = "0.5.1" }
```

near-plugins/Cargo.toml
frolvanya, line 20:
```suggestion
near-plugins-derive = { path = "../near-plugins-derive/", version = "0.5.1" }
```

frolvanya (changes requested):

near-plugins-derive/Cargo.toml
vsavchyn-dev: If we revert dev-dep of `near-plugins-derive` to what was before #188 we will have the same issue once we try releasing `near-plugins-derive` and `near-plugins`.

Current problem in CI from what I see is that `near-plugins-derive` and `near-plugins` have circular dependency. Even though `near-plugins` is only a dev-dep, `cargo-publish` still tries to look for 0.5.2, which doesn't exist during run of the `publish` command. Unpinning the version of `near-plugins` should fix that afaiu.

Here is cargo publish with unpinned version:
image

And cargo publish with pinned:
image

And considering that the order of publishing should be `near-plugins-derive` -> `near-plugins`, we should have things released once we merge release-plz pr.

near-plugins/Cargo.toml
vsavchyn-dev: I guess `near-plugins` and `near-plugins-derive` in deps should stay the same version (0.5.2), no? From what I see those crates should go hand to hand in versions.

It's that or we just downgrade the `./Cargo.toml` version to 0.5.1 as well to trigger proper CI rn.

near-plugins-derive/Cargo.toml
frolvanya: ~~How it worked before?~~

Nvm, I thought that this CI was somehow working in the past with the exact same setup and then suddenly stopped. But I see that this CI never worked in the first place

near-plugins/Cargo.toml
frolvanya: If releaze-plz would be able to figure out by itself, then we can leave it as is, but if not, it would create 0.5.3 bump

frolvanya (approved):

near-plugins/Cargo.toml
r-near: Yeah let's see what happens

r-near (approved):

